PR #238, opened 2023-06-08
Fix PreTeXt link.

gigamonkey:
This seems like the right link?


bnmnetp:
Yep, that was a typo... Thanks for the fix!

PR #239, opened 2023-06-10
This fixes part of the CSAwesome conversion.

gigamonkey:
I'm obviously not super familiar yet with this code or the output we're trying to create but this at least helps create valid XML when converting CSAwesome. I think it's right but I may have missed something.


bnmnetp:
Thanks, this looks right to me.

PR #242, opened 2023-06-12
Make sure elements have ids.

gigamonkey:
I'm less sure this is absolutely necessary than some of my other recent PRs—there may be some other place that this could be fixed but at some point this definitely resolved a problem I was having. And it seems good to give things `id`s.

### DIFF
--- a/bases/rsptx/interactives/runestone/activecode/activecode.py
+++ b/bases/rsptx/interactives/runestone/activecode/activecode.py
@@ -454,6 +454,7 @@ class ActiveCode(RunestoneIdDirective):
             suffix = "\n"
 
         course_name = env.config.html_context["course_id"]
+        self.options["divid"] = f"activecode-{self.options.get('divid')}"
         divid = self.options["divid"]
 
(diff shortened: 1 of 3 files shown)


bnmnetp:
Seems harmless at a minimum and probably helpful.  The less the xml writer has to make up stupid ids the better off we are.

bases/rsptx/interactives/runestone/activecode/activecode.py
```diff
@@ -454,6 +454,7 @@ def run(self):
             suffix = "\n"
 
         course_name = env.config.html_context["course_id"]
+        self.options["divid"] = f"activecode-{self.options.get('divid')}"
```
bnmnetp, line 457: I think this would be actively harmful to existing books with identifiers in our database as it would end up adding activecode in front of their ids.  The base class RunestoneIdDirective sets the divid in options.

gigamonkey:
I'm pretty sure I've run the conversion without this change so I think you're right that it's not needed.

PR #243, opened 2023-06-12
Fix for python 3.11

gigamonkey:
I'm actually confused why this is needed. The `path` method is deprecated as of 3.11 but the description of what it does is the same as in 3.10. But with this change I can run `python -m runestone build` under both 3.10 and 3.11 while without it it explodes under 3.11.


bnmnetp:
Looking at the docs, I think we would be better off to use `.as_file()` rather than `.path()`

gigamonkey:
Possibly so. I think I tried that but it wasn't as easy as dropping in the thing in the docs so I fixed the immediate problem.

bnmnetp:
OK, I'll merge and then see about moving to the as_file.

PR #253, opened 2023-06-21
Fix: Move to a Windows-friendly subprocess approach.

bjones1:
This removed the filtering that was added earlier. I didn't fully understand why the filter only accepted those strings, and also think seeing all output is helpful for a person watching the output. The old filter was:

``` Python
    if "naming to docker.io" in line and "done" in line:
        print(line[line.rindex("/") + 1 :].replace("done", "").strip())
```

### DIFF
--- a/build.py
+++ b/build.py
@@ -1,19 +1,28 @@
 #!/usr/bin/env python3
-
-import os
+#
+# `build.py` - Build the wheels and Docker containers needed for this application
+# ===============================================================================
+#
+# Imports
+# -------
+import asyncio
+import io
 import locale
-from shutil import copyfile
+import os
 import subprocess
 import sys
+from shutil import copyfile
+from typing import Any
 
 # use python-dotenv >= 0.21.0
 from dotenv import load_dotenv
 from rich.console import Console
-from rich.table import Table
 from rich.live import Live
+from rich.table import Table
 from rsptx.cl_utils.core import pushd
-import sh
 
+# Check environment variables
+# ---------------------------
 print("Checking your environment")
 if not os.path.exists(".env"):
     print("No .env file found.  Please copy sample.env to .env and edit it.")
@@ -24,6 +33,7 @@ if "--verbose" in sys.argv:
 else:
     VERBOSE = False
 
+# Per the [docs](https://pypi.org/project/python-dotenv/), load `.env` into environment variables.
 load_dotenv()
 console = Console()
 table = Table(title="Environment Variables")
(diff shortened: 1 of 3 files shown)


bjones1:
That makes sense. Do these message appear often enough so that developers will know that the build is still working?

It looks like you're searching for lines like:

```
#18 naming to docker.io/library/rs-jobe 0.2s done
```

build.py
```diff
@@ -118,35 +130,60 @@
                                 f.write(res.stderr.decode(stdout_err_encoding))
 
 
-# When using subprocess there is no nice way to try (short of threads) to capture the output and send
-# it to the console while also capturing it to a file.  So we use sh instead.
-# https://amoffat.github.io/sh/index.html
-# sh "knows" about all of the commands on your PATH
+# Build Docker containers
```
bnmnetp, line 133: I really think the home for this, as I suggested in Discord is in the `components/rsptx/cl_utils/core.py` file.

I would suggest augmenting the run_command with an optional function that can take the place of the output_stream function.  This makes it nearly functionally equivalent to what we had with sh

bjones1:
(The log from a full build was > 10,000 lines, so that's definitely overwhelming.)

bnmnetp:
Yes, the "naming to" pattern was my best guess at telling when a particular image was complete.  I'm open to other suggestions.

bnmnetp:
Thanks.

One more request.  I know you are a big fan of LP, but I think that we do really need to provide some more traditional API style documentation. See - https://runestone-monorepo.readthedocs.io/en/latest/index.html

Particularly the database module documentation and the rsmanage documentation.  I think this style will help our junior developers know what functions are available and what they do without having to wade through the code.

I personally do not find that the LP documentation we have been producing is helpful unless I already know where to look, and when I do I want to see a description of the function , what its parameters are, and maybe an example of how it is used.  Forcing junior developers to wade through the mix of code and comments is not helpful when all they want to know is how do I add a row to the database.  

I would like us to produce the kind of docs you see in the link above.  Is there a way to harmonize what you do for LP with what sphinx provides us in terms of generating the 

I think these are really two different styles of documentation.  The LP docs are great for people who are doing core development.  But they are not satisfying to someone who just wants to use the Runestone API.  Is there a way we can have harmony?

bnmnetp:
To be clear, this is a separable issue from this, but it came up as I was looking at all of the LP style documentation you added to the code from SO.  But when I tried to add it to the API doc tree, there was nothing there.

bjones1:
Brad, this is a good point. Here's a quick response; I'll start a separate issue so we can continue the conversation.

LP as I've implemented it doesn't produce separate API documentation; this is definitely a problem, especially for junior developers trying to find their way around the source. While it allows developers to add comments about a function/method and its parameters, these don't get excerpted into a nice summary of the API. It does provide good facilities for writing text (such as the excellent `developing.rst` file you created). Its unique strength is documenting inside a function/method, which is also important for junior (and senior) developers.

Traditional API docs (Doxygen/Javadoc/Sphinx API docs/etc.) produces great API docs and decent facilities for writing text. (I think it's easier to compose in my WYSIWYG editor than write text, build it with Sphinx, then check to see if rendered as intended). However, it provides no ability to document inside a function/method, which (IMHO) leads to poor long-term maintainability.

I think we need both for this project, but don't (yet) have a method to do this. Let me think a bit more.

bnmnetp:
already merged locally.

PR #290, opened 2023-06-29
Fix $1 $2

gigamonkey:
One of the more cryptic error message students using CSAwesome get from Runestone is the enigmatic:

```
A server error occurred: $1 $2
```

The main way to trigger that error is by adding a `;` immediately after the class name in an active code exercise. This PR fixes two things: it makes the regex that tries to find the name of the class more forgiving so

```
public class Foo;
{

}
```

will still be understood as trying to write a class named `Foo`.

The second is I've changed the error messages to not contain `$1 $2` since the only place that message is used it is not passed arguments.


bnmnetp:
Seems safe enough to me.

PR #294, opened 2023-06-30
Tidying up sample.env

gigamonkey:
I think `.env` files don't actually support variable expansion so I fixed that and removed a stray `export`. And I added a comment about running `poetry self add poetry-dotenv-plugin` which is actually the real point of this PR.

I don't know if the default values I gave for the four `DBURL` variables are right but they work for me in both the docker composed server and when building the book in the shell created by `poetry shell`.

### DIFF
--- a/sample.env
+++ b/sample.env
@@ -1,48 +1,45 @@
 # Copy this file to .env docker-compose will read and use the values from here.
-# The build.py script is also aware of this file
+# The build.py script is also aware of this file and if you run `poetry self add
+# poetry-dotenv-plugin` when you run `poetry shell` it will also pick up these
+# variables and add them to your environment.
 
 # Where is the runestone app located (inside the container)
 RUNESTONE_PATH = /usr/local/lib/python3.10/site-packages/rsptx/web2py_server/applications/runestone
 
-# Database URLs
-# You may want to have a couple of different environment variables set up that essentially point
-# to the same database.  Therefore we have two variables. DEV_DBURL and DC_DEV_DBURL. The DC variant 
-# should only be set in this file for use with docker-compose.  But if you don't set it and you
-# have a DEV_DBURL variable set in your shell compose will use that as a back up.  In My development
-# environment I have:
-# DEV_DBURL= postgresql://user:pass@localhost/runestone_dev
-# DC_DEV_DBURL=postgresql://runestone:runestone@host.docker.internal/runestone_dev
+# Database URLs: You may want to have a couple of different environment
+# variables set up that essentially point to the same database. Therefore we use
+# one set of variables for use within docker, prefixed with DC_ and another pair
+# for use outside of docker. The DC variants should only be set in this file for
+# use with docker-compose. (If you don't set the DC_ variables, docker compose
+# will fall back to the other pair.)
 
-# Uncomment one or more of these.  If you run a database on the host then the DEV_DBURL uses 
-# a nice docker trick to connect to it.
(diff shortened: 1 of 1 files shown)


bnmnetp:
You are correct... I have WEB2PY_CONFIG set in my config.fish / .bashrc file. So that was masking the problem.

bnmnetp:
I think we should document this in the developing.rst file

gigamonkey:
Do you understand poetry well enough to say whether running `poetry self add poetry-dotenv-plugin` has a global affect or just in the project? If you don't know I can try to figure that out and add something to the docs.

gigamonkey:
Hmmm. Seems like `poetry self add` is doing something to `poetry` itself so I guess it has a global effect. A quick test of creating a new directory with a `pyproject.toml` and a `.env` in it shows that the new `poetry shell` in that dir does indeed slurp in the variables in the `.env` file. And this PR on the `poetry` projects makes it seem like per-project plugins is not a thing yet. 

https://github.com/python-poetry/poetry/pull/5740

So I guess that means we can just recommend the plugin as an easy way to deal with the `rs` project but point out that it is a global choice that maybe you don't want.

docs/source/developing.rst
```diff
@@ -218,10 +218,31 @@ These two sets of variables can be identical, but they are separate because it i
...
-When you are doing development you may want to set these in your login shell, But they can all be set in the ``.env`` file in the top level directory.  This file is read by docker-compose and the values are passed to the containers.  You can also set them in the ``docker-compose.yml`` file but that is not recommended.  The ``.env`` file is also used by the ``build.py`` script to set the environment variables for the docker-compose build.  As of this writing (June 2023) rsmanage does not know about the ``.env`` file so you will have to set them in your login shell if you want to use rsmanage.
-
-
+Variables that are important for the host side are probably best set in your
+login shell environment (such as a .bashrc file) But you can also set them in
+the ``.env`` file and as long as you have a RUNESTONE_PATH set commands like
+``rsmanage`` and ``runestone`` will try to read and use those variables.
```
gigamonkey, line 224: I'm a bit confused by this sentence compared to the disclaimer about `rsmanage` at line 232.

bnmnetp:
I have been adding support for `.env` files directly to `rsmanage` and `runestone` so that if you install them from pypi or to a .local installation of python they will try to find a suitable `.env` file.  I like that we can have poetry read it, but that is only going to help if you are in a poetry shell or if you do `poetry run rsmanage`

docs/source/developing.rst
gigamonkey: To be clear, the thing I was confused about was at 232 where it says:

"As of this writing (June 2023) rsmanage does not know about the ``.env`` file so you will have to
set them in your login shell if you want to use rsmanage."

gigamonkey:
Where does it look for the `.env` if you're running it from pypi? (I assume that means if you did a `pip install runestone`) in `RUNESTONE_PATH`? That seems useful.

Anyway, it seems to me in the context of the docs for developers I'd assume they'd be running from the repo not from something they installed with `pip` so it makes sense to support having everything just work if they _are_ in a poetry shell or using poetry to run the commands. Anyway, the mental model for me of, `cd` into the `rs` directory, run `poetry shell` and then everything Just Works is very appealing.

bnmnetp:
Yes, `RUNESTONE_PATH` is used to find the .env file in the root of the rs repo.

Both  `runestone` and `rsmanage` should be doing this now.

I also like the idea of `poetry shell` just doing it.

PR #302, opened 2023-07-12
Wrap active code buttons for mobile devices

bhoffman0:
The active code buttons don't fit and overflow on smaller mobile screens. 
flex-wrap: wrap; will make them wrap instead of overflow.
Suggestions for the future: shorten the text on the buttons or use icons to make them fit.


bhoffman0:
I just added another css line to fix this squished activity number after the unit test results:
image

PR #322, opened 2023-08-06
New practice mode: self-paced when user loads page

presnick:


components/rsptx/db/crud.py
```diff
@@ -1796,7 +1795,6 @@ async def fetch_one_user_topic_practice(
         & (UserTopicPractice.course_name == user.course_name)
         & (UserTopicPractice.chapter_label == last_page_chapter)
         & (UserTopicPractice.sub_chapter_label == last_page_subchapter)
-        & (UserTopicPractice.question_name == qname)
```
bnmnetp, line 1799: I translated this from ajax.py and I thought it seemed wrong at the time, but Iman told me it was correct.

components/rsptx/validation/schemas.py
```diff
@@ -182,6 +182,9 @@ class LastPageDataIncoming(BaseModelNone):
     last_page_url: str  # = Field(None, alias="lastPageUrl") is the manual way
     course_id: str = Field(alias="course")
     completion_flag: int
+    pageLoad: bool
```
bnmnetp, line 185: This is fine, as long as they are always included on the POST.  If they are missing then the caller will get a 422 invalid request.

bases/rsptx/book_server_api/routers/rslogging.py
```diff
@@ -361,50 +360,74 @@ async def updatelastpage(
...
+                await potentially_change_flashcard(course.base_course, lpd["last_page_chapter"], lpd["last_page_subchapter"], user, tz_offset, remove=True)
+        elif request_data.pageLoad and practice_settings.flashcard_creation_method == 3:
+            # self-paced flashcard creation based on loading a page
+            course = await fetch_course(user.course_name)
+            await potentially_change_flashcard(course.base_course, lpd["last_page_chapter"], lpd["last_page_subchapter"], user, tz_offset, add=True)
 
     return make_json_response(detail="Success")
 
 
-async def add_flashcard(
```
bnmnetp, line 371: I'm definitely confused by this.  potentially_change_flashcard is semantically very different from add_flashcard.  And I wrote add flashcard knowing that when one uses peer instruction we want to add questions from peer instruction that people miss to the practice suite.  And there is another researcher that I'm pretty sure is using add_flashcard.

bnmnetp: With the monorepo -- We should be refactoring functions like this to a new component.  something like components/rsptx/practice

bnmnetp: ``poetry poly add component --name practice`` would create the component for you.

components/rsptx/db/crud.py
presnick: I checked with Iman and understand how it works now; he agreed that it shouldn't take qname as a parameter.

Here's the logic. Each UserTopicPractice instance keeps track of one topic for one user. question_name is the last question practiced on that topic; when the topic is practiced it will go on to the next question associated with that topic.  (Or maybe it's encoding the next one that should be asked, and when the topic is practiced, it is set to be the next one to practice. Either way, it's the same for our purposes...)

When we *create* a UserTopicPractice, we need to set question_name to something. So the code sets it to be the name of the first question associated with the topic. This functino isn't called when we do that.

When we *delete* a UserTopicPractice for a topic, we use this function to look up the dbid of the row, then we delete that row_id.

This function doesn't seem to be referenced anywhere else in the repo. 

If we do end up moving the practice tool over at some point, it could make use of this function. But it will also want a version without the question_name in the query. Because it will be looking up the user, chapter, and sub-chapter, in order to find out what the curent question_name is.

components/rsptx/validation/schemas.py
presnick: Yes, these should always be set. See the new lines 313-315 in user_highlights.js.

bases/rsptx/book_server_api/routers/rslogging.py
presnick: `add_flashcard` doesn't show up anywhere else in the repo. Is that what you mean by another researcher using it?

I think it might help for me to explain the mystery of only one flashcard being added here rather than a new flashcard for each question. The logic, it turns out, is that we add one flashcard (UserTopicPractice instance) for each *topic*, not for each question. 

Some of the comments and var names in the code before created confusion about this; I tried to clean some of them up.

So if there are researchers who've written code to add questions from peer instruction to existing topics (i.e., chapter/subchapter combos), that code should *not* be adding a new UserTopicPractice. It should just associate the questions with the topics. The only subtlety might be that if they add questions to subchapters that have zero built-in questions from the book, then they might need to create one UserTopicPractice (for each student) for that topic, because the current logic doesn't create a UserTopicPractice for subchapters with no practice questions.

If there are people who you think might be working on this and misunderstanding how it's supposed to work, please put them in touch with me.

As for starting a new component, I agree that that would be a good idea. It would make it easier for more things to start being migrated to that component over time if we get it started now. I have a feeling, however, that there will be more to it in adding a new component than just that one command you've given me. At a minimum, updating documentation in a few places. I think if I'm going to do this, we should probably do it together in a pair programming session on zoom.

bnmnetp: You are probably right.  I can already think of a couple of edits to pyproject.toml files that will need to be done.  But I think its worth the effort.  Happy to set up a time to zoom.

PR #332, opened 2023-08-13
Remove html level font-size change in rsptx .less

ascholerChemeketa:
The purged code adds a font-size to the HTML element of any pretext book that makes use of runestone components.

At a bit of a loss for why it is injected into the CSS. Purging it does not appear to break any runestone components.

If it does serve a purpose, there must be a better way to achieve the goal than to change entire page's base font size. Happy to tinker with that if you can point me to what this is trying to do.

### DIFF
--- a/bases/rsptx/interactives/ptxrs-bootstrap.less
+++ b/bases/rsptx/interactives/ptxrs-bootstrap.less
@@ -52,7 +52,3 @@
     --componentBorderColor: #939090;
     --questionBgColor: rgb(23, 85, 93);    
 }
-
-html {
-    font-size: 10pt;
-}


bnmnetp:
@ascholerChemeketa -- Some authors are setting font size with `rem` in custom html.  But there seems to be no rhyme or reason where the base font size is getting set.  So it seemed reasonable to set it to some default so that at least every book would have the same starting point.

I rebuilt several books and this didn't seem to have any ill effects. But maybe it did for your book?

Tagging @cpainterwakefield for more context and maybe ideas on what we should do to make sure everyone's book has the look they want.

cpainterwakefield:
When I wrote my book and made custom stylings for it (a couple of years ago), I was following what I thought was a best practice (from reading online), which is to use rem units for my font sizes.  That depends on the base font size being set to something reasonable.  My styles worked until recently  when everything suddenly went up in size (enough to make my book unusable); I don't know what changed in the Runestone code at that time.

Using a local preview with runestone, the browser dev tools says the base font size is set on the html element in scaffolding.less.  I didn't see anything comparable when viewing my book on the Runestone server.  So adding the base element font size setting was one of the possible fixes I proposed (since it seemed like it would just be restoring behavior that was there previously).

I am by no means a serious front-end developer, so I don't know if there is a better way to set the base font size that doesn't require adding a style on the html element.  Or perhaps best practices have changed and I should do my font sizes differently.

bnmnetp:
@ascholerChemeketa -- I thought that modern browsers converted pt to an appropriate number of px.  px seem crazy to me because a px is not a standard size.  4k monitors, retina monitors, phones, the px density is different everywhere you look.

That said, I don't have strong feelings.  This was precipitated by a user who had way way way beyond the readers digest large print edition size fonts and claimed they could not get them to reduce, thus making the page totally unreadable.

I have no idea how to set the default font size in Safari.  I looked.
I can find something in chrome but it is described in terms of small medium large.

using rem measurements seems like a very good idea, but only if you have some idea what the baseline is.

Feel free to propose a way forward.

cpainterwakefield:
The book where things were too big is my book, practical_db.  I used rem, but it was relative to some setting that existed previously, which was apparently 10px or some such (I was unaware that 16px was standard, and was just setting sizes based on what I was seeing).  If we want to remove the 10px setting, then I will have to update my CSS (just a few sizes to fix, so shouldn't be too hard).

PR #334, opened 2023-08-14
Fix: remove internal scrollbar on course assignment list

bhoffman0:
Now that the course change dropdown moved up, you no longer need the height and scrollbar on the assignment list. Right now, there is a double scroll bar and the scrollbar appears even if there are no assignments.
image
When there are no assignments:
image


bhoffman0:
@gigamonkey @bnmnetp  for review

gigamonkey:
Thank you! I saw that but couldn't tell what it was and never got around to figuring it out. I think this is a good change.

On the mac it's not even obviously a scrollbar when there are no assignments:

2023-08-14 at 6 32 AM

PR #341, opened 2023-08-19
Add navbar item to copy canonical href to clipboard.

gigamonkey:
This PR adds a very useful (to me, anyway) feature to make it easy to copy the canonical link to any page you're looking at in any book just by clicking the link icon I added to the navbar. We could make this an instructor only feature but I think it might be useful for students to who might want to send links to friends in other sections of the same class. So I made it available to everyone.

(I also changed the definition of the `<link rel="canonical">` to inherit the protocol from the current page rather than always being `https` to make it easier to test locally.)

### DIFF
--- a/bases/rsptx/interactives/runestone/common/css/runestone-custom-sphinx-bootstrap.css
+++ b/bases/rsptx/interactives/runestone/common/css/runestone-custom-sphinx-bootstrap.css
@@ -1178,28 +1178,3 @@ margin-right: 3px;
   float:right;
   padding-left: 4px;
 }
-
-/* Give some indication that the click is doing something. */
-
-#canonical_permalink {
-  text-decoration: none !important;
-}
-
-#canonical_permalink:active {
-  transform: scale(0.85);
-}
-
-#canonical_permalink::after {
-  content: " Copied!";
-  display: inline-block;
-  opacity: 0;
-  padding-left: 5px;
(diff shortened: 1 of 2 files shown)


bnmnetp:
A couple of comments:

1.  I don't think this needs to be visible to students.  Only instructors.
2. I'm not sure this is broadly useful enough to warrant its own space in the navbar.  Is there a different place we could add this function?  Part of the 👤 menu?  What about adding it as a menu option to the permalink on the heading for the page?

gigamonkey:
Changed to the user menu. I didn't do anything to hide it from non-instructors. I still think it's legit for a kid to possibly want to share a link with another kid using the same text in a different course.

gigamonkey:
@bnmnetp how do you feel about this now?

bnmnetp:
I worry that it is going to confuse the majority of instructors.  Especially those with only a single section.   I just know I'll get lots of bug reports from people that think it is copying the wrong url. 

Maybe there is a way to turn it on just for power users.

gigamonkey:
I'm okay without it.I have a bookmarklet I wrote that will grab the <link rel="canonical" ...> for any page (not just Runestone) so as long as you add that environment variable so the canonical links don't point to localhost I'll be all set.

bnmnetp:
Yes, the canonical links should be fixed.

PR #344, opened 2023-08-20
Try to stop the progress bars from overflowing.

gigamonkey:
When looking at the actual student progress page for one of my current classes I noticed one of the rows getting weirdly indented.

2023-08-20 at 1 54 PM

Poking around a bit it seems that the problem is the percentages of the preceding row, each rounded to two decimal places as they are to set the widths of each segment of the progress bar, add up to 100.01 and thus overflow the enclosing div. This patch tries to fix it. I couldn't figure out a good way to test it on my local cluster without creating 27 fake students, etc. but the page still at least works.


bnmnetp:
Interesting... Thanks for digging into this.

gigamonkey:
Oh, I should say, I did try changing the width of the last bar to 55.55% and that fixed the formatting problem. So assuming the Python code works the way I intended, I'm confident that it will fix the CSS.

PR #354, opened 2023-09-04
use catch2's catch.hpp

### LINKED ISSUE #352: c++ unittests with catch.hpp broken
**What Course are you in**
thinkcpp / overview

**What Page were you on**
overview: `cpp_units`

https://runestone.academy/ns/books/published/overview/ActiveCode/clangs.html

Just press "Save & Run".

**What is your username**
wrigjl

**Describe the bug**
C++ unittests are broken. catch.hpp uses MINSIGSTKSZ and expects the definition to be a constexpr. The Linux folks have decided to define MINSIGSTKSZ in terms of sysconf() which... is not a constexpr.

This happens when `_GNU_SOURCE` or `_SC_SIGSTKSZ_SOURCE` are defined (which appears to be default today).

**Traceback**

```
In file included from /usr/include/signal.h:328,
                 from /usr/include/catch.hpp:8007,
                 from test.cpp:8:
/usr/include/catch.hpp:10791:58: error: call to non-‘constexpr’ function ‘long int sysconf(int)’
10791 |     static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;
      |                                                          ^~~~~~~~~~~
In file included from /usr/include/x86_64-linux-gnu/bits/sigstksz.h:24,
                 from /usr/include/signal.h:328,
                 from /usr/include/catch.hpp:8007,
                 from test.cpp:8:
/usr/include/unistd.h:640:17: note: ‘long int sysconf(int)’ declared here
  640 | extern long int sysconf (int __name) __THROW;
      |                 ^~~~~~~
In file included from test.cpp:8:
/usr/include/catch.hpp:10850:45: error: size of array ‘altStackMem’ is not an integral constant-expression
10850 |     char FatalConditionHandler::altStackMem[sigStackSize] = {};
      |                                             ^~~~~~~~~~~~```

**Javascript Errors**

Thankfully, this is not javascript.

wrigjl:
I think this fixes the problem in #352. It's not clear to me where catch.hpp comes from in the base image (`dpkg -S` doesn't know about it). I figure we control the horizontal and vertical, so nuke the old one and use the new one.

closes #352


bnmnetp:
The version 1 `catch.hpp` is in `rs/projects/jobe/include`

wrigjl:
> The version 1 `catch.hpp` is in `rs/projects/jobe/include`
> 
> It is copied to /usr/include on line 28 of the Dockerfile

I completely missed that (I was looking for catch.hpp).  Hmm, does it make sense to delete the one in our repo and depend on the distro's catch.hpp from catch2?

bnmnetp:
Yes, I would think so.

Do you have a working test server?  This `ln -s catch2/catch.hpp /usr/include/catch.hpp` looks a little suspect as I'm not sure where apt installs `catch2` compared to the working directory.  An absolute path would be better in any case.

wrigjl:
I don't yet, but I've been doing some test compiles by running the jobe image by hand (I think I'm not far away from a working server). I've done `python3 build.py` a bunch of times, so it -should- just be a matter of composing, etc.

PR #362, opened 2023-09-10
force a redirect logic around lti

mlhess:


bnmnetp:
What is this for?  Is there some special case where you want to use lti but not really?

mlhess:
When in LTI_ONLY mode, users can't login to the book without hitting an LTI link first.  Since users have to be authenticated before seeing content, they have to hit a "login LTI link" then go back and go to the page they want to go to. 

This allows us to keep the LTI authentication part, and then redirect them to a book page after authentication.    

The risk here is an open redirect, but hard coding the path (which has to be set for LTI anyway) and making sure the url does not start with / or http  removes that risk.

presnick:
@bnmnetp : I'm hoping you can approve this so we can avoid forking. We are launching with this version on Coursera soon.

bnmnetp:
I asked Mike if he had used this in production or if it had been tested, and the answer was no to both questions.  I don't have an environment to test LTI changes and I can't have it break for all my users.

bnmnetp (approved):

PR #380, opened 2023-10-04
Update addCaption to suppress PreText captions in runestonebase.js

geoff-cox:
@bnmnetp, is this what you meant by checking for a runestone page to suppress the captions in PreText? I don't have a lot of experience coding in js, so please eyeball the syntax. Thanks.


bnmnetp:
Thanks for this... traveling right now but will try to review it in the next week.

bnmnetp:
So, looks like it will hide the unique ids of questions when you are not using a runestone server, but just building a book as a static webpage with no grading or storage.  

I'll test this a bit more but should be merged by the update this weekend.

bnmnetp:
Thanks!  Finally catching up on some long overdue merges.

PR #389, opened 2023-10-09
minor CSS problem

Alex-Jordan:
I see this when I am in an Assignment page and use the dropdown menu:

Screenshot 2023-10-09 at 4 42 55 PM

where the menu content spills over. If I remove the float CSS as in this PR, it becomes:

Screenshot 2023-10-09 at 4 44 37 PM

I'm not sure if the float CSS was in there for a reason, but I couldn't identify a reason to keep it. (However in testing, I found an issue with this page on small screens that I will report separately.)

This is also me testing the waters for submitting a PR to Runestone. Is it helpful to work this way instead of just reporting the issue? Like is this too small of a thing, and I should gather together other small things to package in one PR? Differnet projects have different customs, and I want to do whatever is helpful for RA.


bnmnetp:
Thanks Alex!  I love small PRs that don't take me too long to understand.

PR #405, opened 2023-10-19
fix 399, add clarification

solijonovam:
**Issue:**
The 'Developing' page was a little unclear, and it was hard to find where to go next.

**Fix:** 
added changes to developing.rst file to make the changes, and added clarification that will help the reader understand what to do next.

**To test:**
Go to Developing page, and see the changes added.

### DIFF
--- a/docs/source/developing.rst
+++ b/docs/source/developing.rst
@@ -24,4 +24,4 @@ To get started, work through the following sections of the documents. Each step 
 
 If you are interested in deploying a production version of Runestone to serve books, begin by setting up a development environment. Once you are confident in that process, the :ref:`moving to production instructions<moving-to-production>` will help you set up a production ready server.
 
-To get started with getting the Runestone code go to the following page :ref:`<get-the-code:>`. 
+To get started with getting the Runestone code go to the following page :ref:`<get-the-code>`. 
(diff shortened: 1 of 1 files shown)


docs/source/developing.rst
```diff
@@ -23,3 +23,5 @@ To get started, work through the following sections of the documents. You will n
 * Build the various Runestone servers and run them
 
 If you are interested in deploying a production version of Runestone to serve books, begin by setting up a development environment. Once you are confident in that process, the :ref:`moving to production instructions<moving-to-production>` will help you set up a production ready server.
+
+To get started with getting the Runestone code go to the following page :ref:`<https://runestone-monorepo.readthedocs.io/en/latest/structure.html>`.
```
bnmnetp, line 27: We should avoid hardcoding the actual URL to the docs and use the format of `:ref:` as is used on line 25 above.

pearcej: @solijonovam This means you should change this to a relative link reference rather than an absolute link reference which would break if the pages were restructured.  Basically, you are being asked to match what was done on line 25, which is a relative reference.

pearcej (commented):
@solijonovam This is still not coreectly relative, and the link text is missing. I recommend matching line 25

pearcej (changes requested):
what is the link text? And, who has been reviewing these before you submit @solijonovam?

Please look carefully at line 25.

solijonovam:
> what is the link text? And, who has been reviewing these before you submit @solijonovam?
> 
> Please look carefully at line 25.

would the link text be **get-the-code**  ?

bnmnetp:
It would be nicer if there was a more reader friendly label before the link that is:

```
:ref:`Friendly Text <xref-label>`
```

docs/source/developing.rst
pearcej, line 1: Exactly!

bnmnetp (approved):

Issue #373, opened 2023-09-23
Restoring WW problems


bnmnetp:
Alex,

Moving this to an issue... so we can track more precisely than email.  Tagging @Tanaquil18 and @mattboelkins

This morning I was able to implement some changes to my side of the webwork code that filled in some deficiencies. The result of which is:

1. no more toggle behavior.  The icon reliably appears and *most* answers are reliably restored.
2. The server is updated with the correct percent correct when the student submits.  This is not new, just confirming that the server side autograder has the right data to do its job correctly.
3. Following longstanding Runestone behavior, the latest student answer is stored in local storage on the browser.  When the student brings up a page, we get the answer from the server.  If the answer from the server indicates that it is correct then we accept the answer from the server.  If the answer from the server is not correct then we check local storage.  If the answer from local storage is correct the server is updated.  If the answer from local storage is *newer* than the server then we update the server.
4. Fill in the blank answers always repopulate correctly.
5. drop down menu answers seem flakey - sometimes it seems they are just lost.  I may not be giving Alex enough data or the right data to restore these....
6. Multiple choice answers behave similarly.  again I may not be giving alex what he needs.

Alex, here is what you have to work with....  Seems likely something is missing.

<img width="556" alt="Screenshot 2023-09-23 at 9 38 57 AM" src="https://github.com/RunestoneInteractive/rs/assets/51115/6db2ac42-06cf-479c-828a-2857355fd755">

Alex-Jordan:
Thanks, I will continue with this with the original problem that Chrissy pointed to and see what more I can either do, or at least diagnose.

bnmnetp:
@Alex-Jordan Anything to report on this one?

Alex-Jordan:
No, but thanks for the reminder. I'll try to look today. Do you recall if this data was from the same exercise Chrissy reported with before?

bnmnetp:
Yes, it was.

Alex-Jordan:
Something is screwy with the WW problem in that example, and it is screwy in a way that overlaps with this issue. So without getting into the details here, I want to get to the bottom of that screwiness before I would continue. I'm 95% certain that the problem has an issue with its PG code, or it is exposing a bug with a PG macro library that it relies on.

[...]

PR #409, opened 2023-10-21
for WeBWorK, store student_value with MC exercises

Alex-Jordan:
This should make it so that with MC questions, RS stores the actual value(s) of the input element(s) with MC exercises, instead of text strings associated with those input elements. 

I have not tested this at all because I don't know how I would go about doing that. Does the code look like it makes sense? If so, then the next thing to do is look at what gets saved for the exercises @Tanaquil18 listed in #373 (near the bottom of the thread at present) and see that this is indeed grabbing "values" and not strings that are unhelpful for identifying which items to select when restoring previous answers.

After this is in, I will have follow-up work in `pretext-webwork.js` to actually make use of what will now be saved. And it should (fingers crossed) put #373 to bed.

### DIFF
--- a/bases/rsptx/interactives/runestone/webwork/js/webwork.js
+++ b/bases/rsptx/interactives/runestone/webwork/js/webwork.js
@@ -96,12 +96,16 @@ class WebWork extends RunestoneBase {
             if (data.rh_result.answers[k].score == 1) {
                 correctCount += 1;
             }
+            // mostly grab original_student_ans, but grab student_value for MC exercises
+            let student_ans = ['Value (parserRadioButtons)', 'Value (PopUp)', 'Value (CheckboxList)'].includes(data.rh_result.answers[k].type)
+                ? data.rh_result.answers[k].student_value
+                : data.rh_result.answers[k].original_student_ans;
             this.answerObj.answers[
                 k
-            ] = `${data.rh_result.answers[k].original_student_ans}`;
+            ] = `${student_ans}`;
             let mqKey = `MaThQuIlL_${k}`;
             this.answerObj.mqAnswers[mqKey] = data.inputs_ref[mqKey];
-            actString += `actual:${data.rh_result.answers[k].original_student_ans}:expected:${data.rh_result.answers[k].correct_value}:`;
+            actString += `actual:${student_ans}:expected:${data.rh_result.answers[k].correct_value}:`;
         }
         let pct = correctCount / qCount;
         // If this.percent is set, then runestonebase will transmit it as part of


Alex-Jordan:
My javascript syntax is often not so great. I had a suspicion after submitting this that I might need parentheses around the trinary operator where `student_ans` is defined. Does that sound right?

bases/rsptx/interactives/runestone/webwork/js/webwork.js
```diff
@@ -96,12 +96,16 @@ class WebWork extends RunestoneBase {
             if (data.rh_result.answers[k].score == 1) {
                 correctCount += 1;
             }
+            // mostly grab original_student_ans, but grab student_value for MC exercises
+            let student_ans = (['Value (parserRadioButtons)', 'Value (PopUp)' 'Value (CheckboxList)'].includes(data.rh_result.answers[k].type))
```
bnmnetp, line 100: I think you are missing a `,` before `'Value (Checkbox...`

Also I think the syntax is wrong.  For the conditional assignment statement operator the syntax is

`var = value_if_true : condition ? value_if_false`

Alex-Jordan: Sorry about the comma, and thanks for catching it.

I think I have the conditional operator right, or at least one way of having it right. [Here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator) it's `condition ? ifTrue : ifFalse`.

I just force pushed with the comma and a little simplification.

bnmnetp: I think that is why I typically just use if statements.

bnmnetp:
OK, so I am trying this out on checkpoint 2.2.3 in the "webwork/sample-chapter" book.

I store:
``` 
{"answers": {"AnSwEr0002": "B1", "AnSwEr0001": "Blue", AnSwEr0003": "B1"}, "mqAnswers": {}}
```
It restores the dropdown correctly but not the radio button or the checkboxes

Alex-Jordan:
It won't restore things properly until I make changes to pretext-webwork.js based on this change here. I suppose I could try to make those changes now to help you see that it all comes out working with your testing. Tomorrow...

bnmnetp:
OK, I was planning to make a release today, to fix some other ongoing problems.

Do you think this is safe to merge in the spirit of "do no harm" ?

Or, I could build that book on `dev.runestoneacademy.org` without merging and you can use that to do your testing?

Alex-Jordan:
I think it is safe. It may do a tiny mount of harm until I get that pretext-webwork.js updated. The harm is that there are some types of MC questions that presently do repopulate answers, but might stop if this is merged and I have not yet made those corresponding updates. IMO it's acceptable. Especially if it only takes me 24 hours or so to get to it.

bnmnetp:
That sounds good.  I'll go ahead and merge.  I'll let you know when you have a book to test with.

bnmnetp:
OK, you can use the PTXWW book on runestone.academy to do some testing.

Alex-Jordan:
All of the sample Multiple Choice exercises currently at PTXWW are restoring answers now. I am unsure that the current MC answers in that book cover the full spectrum of how MC input values are handled in PG problems. 

I went to Chrissy's Active Calculus to try the exercises we've known have this issue, and it is not successfully restoring answers there. But I see evidence that the changes from this PR are not happening there. Do I need to wait for AC to be regenerated before I could test those exercises from this book?

bnmnetp:
Yes!  that is great news.  I will rebuild AC so it has your latest fixes on the Runestone side. I can probably even safely do that on a Friday afternoon.

Alex-Jordan:
Actually I realized that AC is using PTX js 0.31, and I was only editing 0.32. I'm gathering evidence now and maybe things are fine even without a rebuild.

bnmnetp:
AC definitely will not have the change that I merged yesterday until I rebuild.

Alex-Jordan:
OK, that makes sense and agrees with how the testing I just did came out negative :)

bnmnetp:
Rebuilding now, will let you know when it finishes.

bnmnetp:
Build is finished.  I'm in the car so Discord would be easier if you need to talk through anything.

Issue #406, opened 2023-10-19
Possibly wrong commend in documentation


moisedk:
**Please check the current issues** Many bug reports are duplicates, this just creates more work for us. Searching the issues first may give you your answer or a workaround! If not adding new information to an existing report is much more helpful than a new report.

**What Course are you in**
N/A

**What Page were you on**
N/A
**What is your username**
N/A
**Describe the bug**
The command provided `docker compose -f db.compose.yml up -d db` might not be correct. It does not work. The one suggested int the db.compose.yml file `docker compose -f docker-compose.yml -f db.compose.yml up -d db
` works.
**Traceback**
Not  really a stacktrace; this error was thrown when I ran the first command; Sometime `runestone` changes to `assignment`

service "runestone" has neither an image nor a build context specified: invalid compose project

**Javascript Errors**
N/A

PR #421, opened 2023-11-07
[#406] Fix starting database error

moisedk:
[#406] Comment out the lines in db.compose.yml that start other services other than db

### DIFF
--- a/db.compose.yml
+++ b/db.compose.yml
@@ -34,9 +34,9 @@ services:
         ports: 
           - 2345:5432
 
-    runestone:
-      links:
-        - db
+    # runestone:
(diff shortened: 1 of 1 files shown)


bnmnetp (changes requested):
Although this may work fine for the one case where you want to start just the database, this is not appropriate for long term use as it is very important that the other commands know about the db service.  

Your first PR that fixes the command to start the db initially by including `-f docker-compose.yml and -f db.compose.yml` is more appropriate long term solution.

Check the current documentation as I think I have fixed the command.

moisedk:
Okay, I am closing this PR then.

PR #427, opened 2023-11-16
issue#31 Report the language employed by a Scratch ActiveCode Window

### LINKED ISSUE #31: Report the language employed by a Scratch ActiveCode Window
With Scratch ActiveCode allowing multiple possible languages, it might help to inform a reader what language is being interpreted.  I couldn't see it anywhere.

A suggestion.  Change line 127 of `runestone/activecode/js/acfactory.js` to something like

```html
<h4 class="modal-title">Scratch ActiveCode (${lang})</h4>
```

Better would be a dictionary that translates the "shorthand" versions of `lang` into proper language names, with capitalization, such as `cpp` -> "C++".  That felt like too much for me to try to do with a PR that I couldn't test properly, so an issue is going to be as good as it gets.  ;-)

njounkengdaizem:
fix #31 
### Issue 31
With Scratch ActiveCode allowing multiple possible languages, it might help to inform a reader what language is being interpreted. I couldn't see it anywhere.

### My change does the following

- State present language being interpreted. 
e.g for a cpp course --> Scratch ActiveCode (C++)
- Refactor code that sets the stdin string. Now makes use of the languageNames dictionary.  

#### Before

before

#### After 
after

### DIFF
--- a/bases/rsptx/interactives/runestone/activecode/js/acfactory.js
+++ b/bases/rsptx/interactives/runestone/activecode/js/acfactory.js
@@ -118,6 +118,7 @@ export default class ACFactory {
         const languageNames = {
             'cpp': 'C++',
             'c': 'C',
+            'html':'HTML',
             'htmlmixed': 'HTML',
             'java': 'Java',
             'javascript': 'JavaScript',
--- a/bases/rsptx/interactives/runestone/activecode/js/acfactory.js
+++ b/bases/rsptx/interactives/runestone/activecode/js/acfactory.js
@@ -112,16 +112,27 @@ export default class ACFactory {
         var lang = eBookConfig.acDefaultLanguage
             ? eBookConfig.acDefaultLanguage
             : "python";
-
-        const languageNames = {
-            'cpp': 'C++',
-            'java': 'Java',
-            'python3': 'Python 3',
-            'python': 'Python'
-        };
-        if (Object.keys(languageNames).includes(lang)) {
+        if (lang === "java" || lang === "cpp" || lang === "python3") {
             stdin = `data-stdin="text for stdin"`;
         }
+        const languageNames = {
+            'cpp': 'C++',
+            'c': 'C',
+            'htmlmixed': 'HTML',
+            'java': 'Java',
+            'javascript': 'JavaScript',
+            'js':'JavaScript',
+            'octave': 'Octave',
+            'python': 'Python',
+            'py2':'Python 2',
+            'python2': 'Python 2',
+            'py3':'Python 3',
+            'py3anaconda': 'Python 3 with Anaconda',
+            'python3': 'Python 3',
+            'ruby': 'Ruby',
+            'sql': 'SQL',
+            'ts': 'TypeScript'
+        };
 
         // generate the HTML
         var html = `<div class="ptx-runestone-container"><div id="ac_modal_${divid}" class="modal fade">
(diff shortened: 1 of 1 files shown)


bases/rsptx/interactives/runestone/activecode/js/acfactory.js
```diff
@@ -112,25 +112,29 @@ export default class ACFactory {
         var lang = eBookConfig.acDefaultLanguage
             ? eBookConfig.acDefaultLanguage
             : "python";
-        if (lang === "java" || lang === "cpp" || lang === "python3") {
```
bnmnetp, line 115: This statement  does not contain all of the languages supported by runestone.  Only the languages that require the backend JOBE server to compile.  So you will miss languages like javascript, sql, html. But you cannot just add them to the if statement or use the `.includes` statement with all of them.

bnmnetp (changes requested):

bases/rsptx/interactives/runestone/activecode/js/acfactory.js
njounkengdaizem: 👌I wondered... 
I'll go ahead and revert that change then.
@bnmnetp 
Out of curiosity, with respect to my languageNames dictionary, will I have to handle all the other languages supported by runestone? If that is the case, then the dictionary suggestion might have to be extended to handle those as well

bnmnetp: Yes, the dictionary should handle all of the languages we support.

njounkengdaizem: Can you be kind enough to direct me where to find those? I can poke around but your suggestion will make my life easier😂

bnmnetp: I think you can discern most of them by looking at the names in that if statement, plus the files in the js folder.  activecode_xxx.js There may also be a comprehensive list in activecode.js which is the baseclass for all of those.

njounkengdaizem: Those are all the languages I could find. Hopefully, I am not missing any. 
One thing I wasn't sure is if the `eBookConfig.acDefaultLanguage` can ever be abbreviated as shown by this comment below. I decided to handle those cases either ways.

>    //   lang - to render labels in a style appropriate for other languages,
>     //          and to display the proper language in langDisplayDiv:
>     //          'py2' for Python 2, 'py3' for Python 3, 'js' for JavaScript, 'java' for Java,
>     //          'ts' for TypeScript, 'ruby' for Ruby, 'c' for C, 'cpp' for C++
>     //          'py3anaconda' for Python 3 with Anaconda

bnmnetp: Just make sure that you don't crash out if the language is not in the dictionary.  Make a default, like python, but maybe also have your message say 'xxx is a known language please report this'

njounkengdaizem: Hmm, doesn't make sense to default to python if language is not in the dictionary. Its like defaulting to python for `kotlin`, this might just create misunderstanding

 What about defaulting to the value of `lang` instead even if it's not a programming language e.g cpp. Then add that message 'xxx is a known language please report this' for clarity

bnmnetp:
Nice Job.  Thank you for your contribution.  It should be live in all books tomorrow.

PR #428, opened 2023-11-16
I formatted the time stamp.

### LINKED ISSUE #393: Use local time for selecting and displaying due times
**Is your feature request related to a problem? Please describe.**
When selecting the due *time* for an assignment, it is unclear what timezone those are given in.  Similarly, when students look at an assignment, the time given for the due date is not obvious.  

**Describe the solution you'd like**
For instructors: the time you select when making an assignment is shown as a local time (preferably with a timezone code).  In the grader interface, the "Only check work submitted by" message is given in local time (instead of GMT - 600, for example). 

For students, these times are shown in each interface, always as local times in the same format.  This includes on the "Assignments" page that has a summary of all assignments, at the top of each assignment, and in the gradebook.

timalsinab:
Fix #393 

I changed grading.html, added a line break to display the date, and removed the inconsistency under the grading tab on the instructor's page.
image

Reviewed by: @solijonovam

### DIFF
--- a/bases/rsptx/interactives/runestone/shortanswer/js/shortanswer.js
+++ b/bases/rsptx/interactives/runestone/shortanswer/js/shortanswer.js
@@ -228,6 +228,15 @@ export default class ShortAnswer extends RunestoneBase {
             }
         }
     }
+    formatTimestamp(date) {
+        let year = date.getFullYear();
+        let month = ('0' + (date.getMonth() + 1)).slice(-2);
+        let day = ('0' + date.getDate()).slice(-2);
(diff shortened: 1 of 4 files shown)


bnmnetp:
This is a good start.  The issue was referring to making several different places where a timestamp is displayed consistent.

timalsinab:
image

Do you want the same format on all the time stamp? For example : 2023/08/23 11:00:00 on all the places the time stamp is displayed? 

And can you please let us know what file in the codebase has this? 
image

Thanks in advance.

timalsinab:
image
image

PR #454, opened 2023-12-28
fixed missing pretext dependency

presnick:
Had a missing pretext dependency. Not sure if this is where you wanted it fixed, but this got me past the problem in my local environment.


projects/rsmanage/pyproject.toml
```diff
@@ -21,6 +21,7 @@ python = "^3.10"
 asyncclick = "^8.1.3.4"
 redis = "^4.5.4"
 pgcli = "^3.5.0"
+pretext = "^1.0.0"
```
bnmnetp, line 24: This should be "^2.0.0"

presnick: I think I copied it from projects/w2p_login_assign_grade/pyproject.toml

OK by me to reject this PR and harmonize as you see fit.

bnmnetp: You picked the wrong example 😄   Will close and get everything synced momentarily.

PR #458, opened 2024-01-10
Add new question type for DoenetML

jaltekruse:
This changeset adds a new question type for DoenetML activities.

The code changes were based loosely on the hparsons question type, and there are some remaining artifacts from having copy-pasted code from those changes that are lingering, I should be cleaning those up soon.

I have added code the question authoring interface in the assignment editor to enable using the DoenetML 2 panel editor that can generate previews of question content all on the client without a trip to the server.

I would like to remove the need to include the question id in the source code of the question, right now it gets stuck in an XML comment, and I generate a random number automatically for authors.

I should probably open an issue to discuss that, because I would like to improve the general UX of authors needing to come up with their own unique keys for question ids if they choose to customize a question.

There are a few enhancements and cleanups that I still want to make to this work, but I wanted to get it posted so others could start reviewing it.

One thing that is not in it's final state is how I am referencing the built version of Doenet. I just have a hard-coded URL to a random personal repo I can push updates to easily. In the production version I think we likely want to put this artifact on the doenet.org server, or maybe some CDN service. We could consider pulling in Doenet as an NPM dependency (we haven't published this yet but we do intend to do it at some point) and integrating it with the runestone-interactives, but I'm thinking it might be better to allow those two things to live more independently and for us to be able to push out updates without needing to make the Runestone community approve a new release.


jaltekruse:
One important note, this should not be merged as is, it breaks the default included MathJax in the other question types, I still need to fix that.

bnmnetp:
Yes, not good.  I've tested MathJax 4 and that appears to work just fine with all of the other components.

bases/rsptx/interactives/runestone/doenet/doenet.py
```diff
@@ -0,0 +1,181 @@
...
+)
+from runestone.common.runestonedirective import (
+    RunestoneIdDirective,
+    RunestoneIdNode,
+)
+
+
+def setup(app):
+    app.add_directive("doenet", DoenetDirective)
+    app.add_node(DoenetNode, html=(visit_hp_html, depart_hp_html))
```
dqnykamp, line 38: For clarity, should probably rename all "hp" and "hparsons" references.

PR #462, opened 2024-01-18
Add dev environment that removes the need to rebuild containers

jaltekruse:
- The dstart script was already present, but only started individual servers. It is hard to work with because different URLs are served by different microservices, which are served on different ports and clicking links wouldn't lead to the right address. Some individual pages require APIs or static files served from different services so cannot render properly at all.
- Add new nginx container called nginx_dstart_dev that can be used to navigate the site at localhost:8080 and redirects requests to the host machine python processes on the ports specified in the dstart script
- add option "all" to the dstart script to start all of the servers in the background
- add option "stopall" to the dstart script to stop all of the servers, as the "all" option starts them in the background so they are harder to find/kill manually

### DIFF
--- a/dstart
+++ b/dstart
@@ -10,6 +10,13 @@ fi
 
 set -e
 SERVICE=$1
+
+# these environment variables need to be set differentlty for running the servers on the host machine
+# instead of in the docker container, so fix them up here
+
+export RUNESTONE_PATH=.
+export BOOK_PATH=~/Runestone/books
+
 cd $RUNESTONE_PATH
 
 # check to see if the rs virtual environment is active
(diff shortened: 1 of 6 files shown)


bnmnetp:
Just looking at this a bit more closely.  In the nginx_dstart_dev folder the Dockerfile COPYs the `components/rsptx/templates/staticAssets` It would be better to mount that otherwise it will require a rebuild to get any changes to the templates or javascript that lives under the staticAssets folder.

dstart
```diff
@@ -10,6 +10,13 @@ fi
...
 set -e
 SERVICE=$1
+
+# these environment variables need to be set differentlty for running the servers on the host machine
+# instead of in the docker container, so fix them up here
+
+export RUNESTONE_PATH=.
+export BOOK_PATH=~/Runestone/books
+
 cd $RUNESTONE_PATH
```
jaltekruse, line 20: @bnmnetp I just wanted to call this out quickly, the lines I added above help me if I always run this script from my root rs folder, but they make this line a no-op. I don't know if you have use cases for running this script when your current working directory isn't the project root, and/or if you have a different way you would prefer to solve this issue of needing different paths in the 2 contexts (I know I'm setting BOOK_PATH to the same location as the default .env file, but I was seeing that somewhere it was trying to resolve it wasn't able to interpret ~ correctly. Doing it this way solves it I assume because bash resolves it and puts the full path in the environment variable).

bnmnetp: I think it works fine for me because

1. I always run dstart from the rs root.
2. I have both `RUNESTONE_PATH` and `BOOK_PATH` set in my fish config file.

bnmnetp:
The `docker compose --profile dev up -d nginx_dstart_dev` could also be added to the "all" startup in `dstart`

jaltekruse:
> Just looking at this a bit more closely. In the nginx_dstart_dev folder the Dockerfile COPYs the `components/rsptx/templates/staticAssets` It would be better to mount that otherwise it will require a rebu

Done

PR #464, opened 2024-01-23
Compose profiles

bnmnetp:
This uses some of the newer features of docker compose to simplify our setup.

With this, a first time user would not need anything more than the docker-compose.yml and the default settings provided in sample.env


bnmnetp:
TODO:

- [ ] Update the docs
- [ ] Add `nginx_dstart_dev` service with a `dev` profile once I merge #462

bnmnetp:
@jaltekruse @pearcej @ascholerChemeketa 

This seems much cleaner to me **and** easier to set up a single environment variable `COMPOSE_PROFILES` which is default to basic in the sample.env file.

Any thoughts?

pearcej:
Much easier!  Thanks @bnmnetp !

ascholerChemeketa:
Makes sense to me!

PR #474, opened 2024-03-18
Enforce 0 as min score for lineGrader Parsons

ascholerChemeketa:
World's dumbest PR...

A better approach would be to only penalize incorrectly indented correct lines. But that would be a significant refactor for very little gain.


bnmnetp:
LOL...

Issue #478, opened 2024-04-15
Grading short answer questions that involve MathJax slow to load 


ascholerChemeketa:
In the instructor grading interface, when you load 1+ short answer questions for 1+ students, the work takes a long time to load.

It looks like MathJax gets called too many times to format the resulting HTML.

ascholerChemeketa:
Research:

renderRunestoneComponent called for each question/student pair. It does:
`MathJax.typeset([document.querySelector(`#${whereDiv}`)]);` which renders mathjax in the question text.

renderHTML calls `this.queueMathJax(this.containerDiv);` which tries to rerender the entire question container (prompt, student output, etc...)

restoreAnswers calls renderMath on the the answer text which does `this.queueMathJax(this.renderedAnswer)`
    
queued up mathJax calls end up going through:
```
            let payload = await MathJax.startup
                .defaultPageReady()
                .then(async function () {
                    console.log(
                        `MathJax Ready -- dequeing a typesetting run for ${item.component.id}`
                    );
                    return await MathJax.typesetPromise([item.component]);
                });
```

In that, `MathJax.startup.defaultPageReady()` ends up calling `MathJax.typesetPromise(e.CONFIG.elements)` where `e.CONFIG.elements` is null.

So each queue call ends up triggering two typeset promises, one with the component and one with no params.

bnmnetp:
I'm not sure I'm following all of that, but it sounds messier than it should be.

Some quick initial thoughts/questions
1. If the `e.CONFIG.elements` is null does that end up with mathjax trying to re-render everything on the page?
2. I **think** the call to `MathJax.typeset` in `renderRunestoneComponent` can be removed.  I think it was a stopgap that I put in years ago when we first integrated with PreTeXt.  
3. Each component should/does/needs to handle its own MathJax renders.
4. Later versions of mathjax added the async promised based interface as the preferred way of doing things.
5. I'm totally unclear as to why queueing a request would cause an extra call to `MathJax.typesetPromise(e.CONFIG.elements)`

bnmnetp:
It *looks* like in the startup block for mathjax we could set `elements` to an empty list for the grading page rather than the default of null.

bnmnetp:
I think my assumption was that `MathJax.startup.defaultPageReady()` would be a no-op after it was called the first time.   If that is a bad assumption then we should use the `runestoneMathReady` promise rather than a call to that function in the `queueMathJax` function.

bnmnetp:
[...]

PR #480, opened 2024-04-21
Prune excess mathjax when grading questions

ascholerChemeketa:
Addresses issue #478 

Reuse `window.runestoneMathReady` promise instead of triggering a new `MathJax.startup.defaultPageReady()` promise.

Remove old stopgap `MathJax.typeset` in admin.js renderComponent.

Seems to still render fine, reduces the number of mathjax calls and avoids calls to typesetPromise with null element list.


bnmnetp:
Thanks @ascholerChemeketa I'll give this a test locally with the sample book later today and merge.

PR #482, opened 2024-04-27
Added changes for local file storage

aryank-prog:
This JavaScript implementation provides a basic caching mechanism using local storage. It checks for cached data before fetching it from a server, reducing load and speeding up data retrieval.

Functions
fetchDataFromServer(url): Fetches data from a given URL and returns it as a string. It uses the Fetch API.

storeDataInLocalStorage(key, data): Stores data along with a timestamp in local storage. The timestamp helps in determining the freshness of the data.

retrieveDataFromLocalStorage(key, expiry): Retrieves data from local storage if it is within the freshness threshold defined by expiry. If the data is stale, it is automatically removed.

getData(url, key, expiry): Orchestrates the process of retrieving data. It first checks for fresh data in the cache. If none is available, it fetches data from the specified URL and caches it for future use.


bnmnetp:

PR #483, opened 2024-04-27
Grading fix branch fix

aryank-prog:
This HTML document creates a webpage with a flexible two-column layout. The layout adapts to the screen size, using CSS Flexbox to ensure that columns do not overlap and are displayed effectively across devices. The design is responsive, with columns that stack vertically on smaller screens.


bnmnetp:

PR #484, opened 2024-04-27
Retrievinglocalstorage

aryank-prog:
This implementation provides a mechanism for retrieving a datafile by first checking for a local copy stored in the session storage. It prioritizes using the local copy if it is fresh to reduce server load and improve response times.

Functions
getDataFromSessionStorage(key): Checks session storage for a datafile using the provided key. It ensures the data is fresh based on a predefined freshness limit (1 hour).

storeDataInSessionStorage(key, data): Stores a datafile in session storage with the current timestamp to track freshness.

fetchDataFile(url, key): Manages the retrieval of a datafile. It first attempts to use a fresh local copy from session storage. If not available, it fetches the data from a server and stores it locally for future use.


bnmnetp:
Why are you making PRs with nothing but empty files?

bnmnetp:

PR #497, opened 2024-07-07
Components dev improvements

ascholerChemeketa:
Improves workflow for Runestone js development with PreTeXt books.

Goes with https://github.com/PreTeXtBook/pretext/pull/2197

### DIFF
--- a/bases/rsptx/interactives/webpack.config.js
+++ b/bases/rsptx/interactives/webpack.config.js
@@ -37,7 +37,7 @@ function definePluginDict(env) {
 module.exports = (env, argv) => {
     const is_dev_mode = argv.mode === "development";
     // ``env.builddir`` can be used to specify the output directory
-    // Example: ``npm run build -- --env builddir=..\mypretextbook\output\_static``
+    // Example: ``npm run build -- --env builddir=../mypretextbook/output/_static``
     const dest_path = env.builddir == undefined ?
             path.resolve(__dirname, "runestone/dist") :
             path.resolve(env.builddir);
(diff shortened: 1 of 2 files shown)


bnmnetp:
The last time I merged anything here was in Tacoma.  Seems odd, but we might have to talk through them.

ascholerChemeketa:
Updated to assume RS files will be in `PRETEXT_OUTPUT/_static` without a subdir.

bnmnetp:
I'm trying this out.  I'm getting the files into the right directory but I'm still seeing
```
vendors-node_modules_handsontable_es_index_js-node_modules_handsontable_dist_handsontable_ful-dcc440.js
```
I was thinking that would be gone.  We are also missing the webpack_static_imports.xml file.  Only the json file is present.

Maybe another extra command

ascholerChemeketa:
The only `vendors-node_modules_...` files that get renamed are the entry files (js and CSS). They are the ones that are directly loaded by the webpage and thus need steady names. So we get `prefix-runtime-libs.bundle.js` and `prefix-runtime-libs.css` instead of the possibly changing `vendor-node_modules_....` names for the entry file for the libs.

The other `vendors-node_modules_...` files are chunks that webpack will load from the entry points. Those do not need cleaned-up names - I'm pretty sure trying to rename them would require all kinds of fixes to the webpack chunk loading.

------

I don't think developers need to output `webpack_static_imports.xml` - that isn't currently part of the webpack workflow. It looks like you make that with a separate shell script. Taking that part up with you in discord...

bnmnetp:
So we will need to assume that the pretext book **has** been built with the `webpack_static_imports.xml` once and then we can do our dev builds of the javascript knowing that the pages have the right references in them.

bnmnetp (approved):

PR #500, opened 2024-07-10
Executable parsons

ascholerChemeketa:
This goes with pretext PR: https://github.com/PreTeXtBook/pretext/pull/2204

I am fairly confident about RS side. PTX side may need some tweaks.

### DIFF
--- a/bases/rsptx/interactives/runestone/parsons/js/parsons-i18n.en.js
+++ b/bases/rsptx/interactives/runestone/parsons/js/parsons-i18n.en.js
@@ -10,6 +10,8 @@ $.i18n().load({
             "Perfect!  It took you only one try to solve this.  Great job!",
         msg_parson_correct:
             "Perfect!  It took you $1 tries to solve this.  Click Reset to try to solve it in one attempt.",
+        msg_parson_correct_runnable:
+            "Click Run below to test out your solution.",
         msg_parson_wrong_indent:
             "This block is not indented correctly. Either indent it more by dragging it right or reduce the indention by dragging it left.",
         msg_parson_wrong_indents:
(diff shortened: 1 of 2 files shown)


ascholerChemeketa:
@bnmnetp PTX side is merged but non-functional until this JS is merged.

@rbeezer if you want to test the functionality of the runnable Parsons, you can use the `debug.rs.dev` flag on the Sample Book and then grab this PR and build it. Instructions for building local Runestone components are here:
https://github.com/RunestoneInteractive/rs/blob/main/docs/source/javascript_feature.rst

bnmnetp:
@ascholerChemeketa looks good.  
I made the primes example runnable in the sample book.
The only issue I notice is that the horizontal output does not scroll, like it does in other cases.  The sieve program outputs one very long line and it ends up making the editor window wider and the stdout wider also... 

I'm fine merging this and fixing that in a followup unless you want to fix it  in this PR.

bnmnetp:
@rbeezer this is merged, working my way through the accumulation.

bnmnetp:
Question for @rbeezer and @ascholerChemeketa Does the `debug.rs.dev` replace the `debug.rs.services.file` string param?  I guess technically they could have different uses?

rbeezer:
>  Does the debug.rs.dev replace the debug.rs.services.file

I think tat's for @ascholerChemeketa.  Can't say I tracked everthing too carefully with the relevant PreTeXt PR.

ascholerChemeketa:
I agree they have different uses. `debug.rs.dev` = let me use my development copy of RS. `debug.rs.services.file` = let me test a particular version of RS from the CDN.

There is overlap where a RS dev uses `debug.rs.services.file` to test out an old release. They could instead revert their RS code and use `debug.rs.dev`. Not sure which approach you would use @bnmnetp 

It feels like `debug.rs.services.file` also has a use case where an author pins to an older version of RS temporarily because of some issue they are encountering with the latest code.

I suppose they could be compacted into one construct. If anyone feels strongly about that happening I can PR it.

bnmnetp:
Thanks @ascholerChemeketa -- I think my head was deep down the debugging and testing hole.

PR #503, opened 2024-07-11
Workingon466

jjjonesj:
I made changes to the instructor.py by adding a hidden datetime object and switching the visible boolean to datetime under the new_Assignment function.


bnmnetp:
We have talked about this before...  You cannot remove the `visible` boolean or it will cause thousands of assignments to crash.  In a live system backward compatibility is very important.

You and @creativecoder need to work together and give me one PR to look at.  I can't try to look at two PRs trying to solve the same problem and figure out what is going on.  There are lots of well documented workflows that you can find that will help you do this.  I even suggested as much on @creativecoder 's  PR.

Tagging @pearcej

PR #504, opened 2024-07-11
Fitb update

dbrianwalton:
Javascript-based testing of answers was not fully compatible with string/number testing. Non-dynamic problems with mathematics in the statement were not rendering in MathJax.

Modifications made to adjust the answer testing for the missed cases. Moved the queueing of MathJax rending out of renderDynamicContent to each reference where the html was being updated.

Also btm-expressions (npm webpack package for math objects) had some bug-fixes and improvements that need to be available for math fitb problems.


bnmnetp:
Thanks @dbrianwalton I'll try to look at this and get it merged tomorrow.

PR #510, opened 2024-07-23
Add: Review page for Peer Instruction assignments

ibrahimmoazzam:
This PR contains a Review page for Peer Instruction assignments, meant to help instructors with the following features:

- Shows the questions contained within a Peer Instruction assignment
- Renders the feedback and correct choice for the questions
- Allows instructors to view the questions within the book
- Aggregates the peer votes and creates histograms for both voting stages showing the percentage of votes per choice
- Calculates and shows the Question Effectiveness and Normalized Change (NC) for each question


bnmnetp:
Could you maybe git restore all the the poetry.lock files except those for the assignment server?   I know this is because of my build.py script which updated them for you.  I've since changed that behavior.

I've been dealing with a lot of dependency issues lately and don't want to cause problems.

ibrahimmoazzam:
Done! I was able to revert all the poetry.lock files except those for the assignment server.

bnmnetp:
I have your code on my development machine.  I just tried to bring up a report on a peer assignment I've used many times for testing with a couple of fake students.

Nothing on the report and in the console I get:

```
TypeError: Cannot read properties of null (reading 'length')
    at instructor.js:126:78
    at Array.forEach (<anonymous>)
    at HTMLDocument.<anonymous> (instructor.js:124:19)
```

ibrahimmoazzam:
I've just added a new commit that deals with a few edge cases and TypeErrors; could you please confirm if you're still getting the same error? Everything seems to be working perfectly fine on my machine with no errors in the console. Thank you!

PR #520, opened 2024-08-01
Updates to support FITB/math problems

dbrianwalton:
* Add character indicators (checkmark/x-mark) to FITB feedback for correct/incorrect.
* npm version change for btm-expressions (math formula support)
  * Added ability for formulas to do some automatic algebra reductions
  * Automatic simplification of formulas involving constants

The authored feedback for fillin problems doubles as text for the automatically-generated solution in PreTeXt sources. That feedback does not necessarily have enough context clues to indicate if an answer was correct or not (in the case of multiple blanks), so the addition of a checkmark when correct and and "x" when incorrect should improve visual clues of correctness. The characters match what is used for multiple choice feedback.

The btm-expressions version update facilitates improved automatic simplification of expressions that are used for formulas that are derived from authored formulas so that they can appear cleaner when appearing in solutions.

(Note: I don't know if there is a better way for the btm-expressions library to update. The choice of including it as a Runestone dependency was so that problems always have the resources if the Runestone Component itself loads. But then incremental improvements to the underlying library require an update of package.json without any corresponding actual change to the components. Let me know if we should transition to a different mode.)

### DIFF
--- a/bases/rsptx/interactives/package.json
+++ b/bases/rsptx/interactives/package.json
@@ -37,7 +37,7 @@
         "-": "0.0.1",
         "@mapbox/node-pre-gyp": "^1.0.11",
         "bootstrap": "3.4.1",
-        "btm-expressions": "^0.1.9",
+        "btm-expressions": "^0.1.10",
         "byte-base64": "^1.1.0",
         "codemirror": "^5.59.4",
         "handsontable": "7.2.2",
--- a/bases/rsptx/interactives/package.json
+++ b/bases/rsptx/interactives/package.json
@@ -37,7 +37,7 @@
         "-": "0.0.1",
         "@mapbox/node-pre-gyp": "^1.0.11",
         "bootstrap": "3.4.1",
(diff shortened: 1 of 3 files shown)


dbrianwalton:
Saw a math reduction that wasn't working so I'll need to make another change to the btm-expressions library before this goes up. Shouldn't be long.

dbrianwalton:
Okay. The additional case that I'd missed is now included. Should be ready to go.

bases/rsptx/interactives/package.json
```diff
@@ -37,7 +37,7 @@
         "-": "0.0.1",
         "@mapbox/node-pre-gyp": "^1.0.11",
         "bootstrap": "3.4.1",
-        "btm-expressions": "^0.1.9",
```
bnmnetp, line 40: I think you can just make this "^0.1.0" and it will do 0.1.X automatically.

bnmnetp (approved):

PR #538, opened 2024-09-14
Pass args instead of null

DavidCooperWCU:
This pass through method wasn’t passing necessary args for methods that take arguments. Now it does.

### DIFF
--- a/projects/jobe/files/CodeTestHelper.java
+++ b/projects/jobe/files/CodeTestHelper.java
@@ -439,7 +439,7 @@ public class CodeTestHelper {
                 if (m.getName().equals(methodName)) {
 
                     if (!checkStaticMethod(m) && checkReturnType(m, "void")) {
-                        return getInstanceMethodOutput(m, null);
+                        return getInstanceMethodOutput(m, args);
                     } else if (!checkStaticMethod(m)) {
                         Object o = getTestInstance();
 


bnmnetp:
@DavidCooperWCU 

What issue or problems does this PR address?

@bhoffman0 could your or Kate take a look at this and let me know what you think?

bhoffman0:
I'd like to get @kmcdonnell2 to look at it too, since it's her code.  This is discussed more here https://discord.com/channels/1013815439161315348/1283143831067295766/1283874612572065927.

DavidCooperWCU:
> @DavidCooperWCU 
> 
> 
> 
> What issue or problems does this PR address?
> 

When trying to get method output for instance methods that take arguments, without this fix, the codetesthelper will fail because the arguments are not sent to the method that requires arguments.

bnmnetp:
OK, this looks pretty innocuous, have you tested it ?  Hopefully @kmcdonnell2 can give it a quick spin.  Or maybe there is any easy way to test it on replit or something that is documented somewhere?

DavidCooperWCU:
This is a working test case. Tested on JGrasp, which is why the test class file is not RunestoneTest.java

[WorkingExample.zip](https://github.com/user-attachments/files/17019640/WorkingExample.zip)

kmcdonnell2:
Yes, I believe this change works. I was able to do a quick verification on Repl.it.
https://replit.com/@kmcdonnell2/testCode

bnmnetp:
OK, I will merge, and this should be live with the next update to the site.  Saturday at the latest.

PR #541, opened 2024-09-26
Fix: Interactivity + placement of PI buttons

ibrahimmoazzam:
I have reviewed the new interactivity + arrangement of buttons with Barb and she believes they're better than the existing design.

### DIFF
--- a/bases/rsptx/web2py_server/applications/runestone/static/js/peer.js
+++ b/bases/rsptx/web2py_server/applications/runestone/static/js/peer.js
@@ -301,15 +301,10 @@ function warnAndStopVote(event) {
         let butt = document.querySelector("#vote1");
         butt.classList.replace("btn-info", "btn-secondary");
         document.querySelector("#makep").disabled = false;
-        document.querySelector("#vote2").disabled = false;
         document.querySelector("#facechat").disabled = false;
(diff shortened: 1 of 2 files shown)


bnmnetp:
How about a screen shot so I know what this is about??

ibrahimmoazzam:
This PR addresses the following changes:
- The placement of the buttons to control PI is now in a stacked row format, as shown in the screenshot. This makes it easier for instructors to walk through the logical progression of PI.
- The "Next Question" button is always enabled now as instructors have wanted an option to skip questions without having to go through the voting stages
- "Start Vote 2" is only enabled once an instructor chooses a chat modality (text or in-person chat)
- Choosing one of the chat modalities automatically disabled the other one

Screenshot 2024-09-26 at 2 56 04 PM

bnmnetp:
OK, my only worry is that Next Question should be disabled when the instructor gets to the last question in the assignment.  I just fixed that a week ago, and I want to be sure we don't regress.

ibrahimmoazzam:
Yes, I completely understand. I haven't changed any of that code and I made sure to check that Next Question is disabled when the instructor gets to the last question in an assignment. Hope it helps!

PR #548, opened 2024-10-09
handle long parsons lines without wrap

ascholerChemeketa:
Currently, long code lines in Parsons get broken at whitespace. That is quite confusing where code is involved.

This prevents wrapping code in Parsons. Instead, if necessary, the text inside the block can be scrolled side to side to see the tail end of really long lines.

### DIFF
--- a/bases/rsptx/interactives/runestone/parsons/css/parsons.css
+++ b/bases/rsptx/interactives/runestone/parsons/css/parsons.css
@@ -202,6 +202,7 @@
     clear: both;
     float: left;
     background-color: transparent;
+    white-space: pre; /* prevent wrap */
 }
 
 .parsons .block p {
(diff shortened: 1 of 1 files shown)


ascholerChemeketa:
@bnmnetp I could see someone wanting line wrap for natural language. But for doing any kind of code, or even pseudocode, wrap feels really bad.

I could write the CSS to try to only affect known languages (`preteprint lang-XXX` applied), but that would miss pseudocode.

I think the lesser of two evils is to force people to manually manage linebreaks when authoring.

ascholerChemeketa:
@ascholerChemeketa will check against math in sample book

ascholerChemeketa:
Closing this ... 

Found a deeper issue and am fixing that instead:
https://github.com/RunestoneInteractive/rs/pull/551

PR #553, opened 2024-10-15
Fix link to Discord on README.rst

skiadas:
There was a second set of backticks in one of the links that was somehow missed last week when other links were fixed.


bnmnetp (approved):

bnmnetp:
Thanks, the single versus double backticks are one of the things I like least about RST.

PR #565, opened 2024-10-28
Add doctest, remove catch precompilation

ascholerChemeketa:
@bnmnetp If/when you merge could you update the RS server with doctest.h?


bnmnetp:
@ascholerChemeketa:  It will update when I rebuild and push out new images on Saturday.

PR #569, opened 2024-11-06
Update Jobe dockerfile to match current JobeInABox

ascholerChemeketa:
Not needed, but seems worthwhile to update dockerfile while updating Jobe itself.

### DIFF
--- a/projects/jobe/Dockerfile
+++ b/projects/jobe/Dockerfile
@@ -2,7 +2,7 @@
 # With thanks to David Bowes (d.h.bowes@lancaster.ac.uk) who did all the hard work
 # on this originally.
 
-FROM docker.io/ubuntu:22.04
+FROM docker.io/ubuntu:24.04
 
 # https://github.com/opencontainers/image-spec/blob/master/annotations.md
 LABEL \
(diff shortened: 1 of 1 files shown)


projects/jobe/Dockerfile
```diff
@@ -12,59 +12,58 @@ LABEL \
...
+RUN --mount=type=secret,id=api_keys \
+    export API_KEYS=`cat /run/secrets/api_keys | tr '\n' ' '` && \
+    ln -snf /usr/share/zoneinfo/"$TZ" /etc/localtime && \
     echo "$TZ" > /etc/timezone && \
     apt-get update && \
     apt-get --no-install-recommends install -yq \
-    acl \
-    apache2 \
-    build-essential \
-    catch2 \
```
bnmnetp, line 48: Do we want to keep this for backward compatibility?  I think we do.

projects/jobe/Dockerfile
```diff
@@ -12,59 +12,58 @@ LABEL \
...
-    python3 \
-    python3-pip \
-    python3-setuptools \
-    sqlite3 \
-    sudo \
-    tzdata \
-    unzip \
-    vim \
-    wget && \
-    python3 -m pip install pylint beautifulsoup4 pandas requests altair && \
```
bnmnetp, line 67: definitely need to keep these as some books rely on them.

projects/jobe/Dockerfile
```diff
@@ -12,59 +12,58 @@ LABEL \
     org.opencontainers.image.documentation="https://github.com/trampgeek/jobeinabox" \
     org.opencontainers.image.source="https://github.com/trampgeek/jobeinabox"
 
-ARG TZ=Etc/GMT-0
```
bnmnetp, line 15: I think we want to be in GMT so jobe is in the same timezone as the rest of the servers

projects/jobe/Dockerfile
```diff
@@ -12,59 +12,58 @@ LABEL \
...
-    acl \
-    apache2 \
-    build-essential \
-    catch2 \
-    fp-compiler \
-    git \
-    libapache2-mod-php \
-    nodejs \
-    octave \
-    openjdk-18-jdk \
```
bnmnetp, line 54: I wonder which version of java we end up with if we use default-jdk???  This could be bad for CSAwesome.

bnmnetp (changes requested):

ascholerChemeketa:
Sorry, that was a sloppy end of the day job.

Started a clean redo and am not 100% sure we want the API keys or other changes. Leaving well-enough alone until if/when we go to jobe 2.x

PR #583, opened 2024-11-21
Add: Interval to the PI Review Page to Wait for Question Data

ibrahimmoazzam:
This PR contains some changes related to adding an interval to the PI Review Page since some instructors have said they can't view any data (e.g., the correct answer, feedback, voting results) for certain questions. As I can't replicate the problem due to the Review Page functioning perfectly on my local system, I might have to add more changes to fix the problem.


bnmnetp:
If the data is not in `#questions-data` when the page is loaded, its not going to appear later.   Maybe you want to change to get the data you need using an API, that way you could call the API after a small delay, or call it again if you don't get the data the first time.

I think this would also be more in line with future directions if we should convert all of the peer instruction interfaces to use React.

PR #599, opened 2024-12-16
RUN-21 Improve UX when adding/removing readings/exercises

morozov-av:
@bnmnetp 

- The "Add Exercise" button is now sticky.
- All actions related to adding/removing readings/exercises now display a Toast message to make the process more user-friendly.


morozov-av:
@bnmnetp Updated the PR: added popups for tab switching and fixed the bug with incorrect display of the selected exercises count. Please take a look and share your feedback.

PR #619, opened 2024-12-31
RUN-20/RUN-25 Autoupdate assignment points when adding or removing exercise/The assignment type of Regular should be selected by default

morozov-av:


bases/rsptx/assignment_server_api/assignment_builder/src/renderers/assignment.jsx
```diff
@@ -76,48 +77,48 @@ function diff(oldobj, newobj) {
...
-//         }
-//       }
-//     } else {
-//       //console.log("no changes")
-//     }
-//   }, 1500);
-// }
-
-//const unsubscribe = store.subscribe(handleChange); // eslint-disable-line
+function handleChange() {
```
bnmnetp, line 80: Interesting....  I always felt that there must be a better way to achieve this...maybe something we can talk about later.

bases/rsptx/assignment_server_api/routers/instructor.py
```diff
@@ -66,6 +69,66 @@
...
+                )
+
+            user_is_instructor = await is_instructor(request, user=user)
+            if not user_is_instructor:
+                raise HTTPException(
+                    status_code=status.HTTP_401_UNAUTHORIZED,
+                    detail="User is not an instructor",
+                )
+
+            # Continue execution of the original function
```
bnmnetp, line 93: Some functions will require the user for more than just instructor validation.  Maybe we want to inject it into kwargs if the "user" key is already there?  Or maybe the regular dependency injection will work fine.

Not a worry for this PR, but something I want to think about.  Especially if this is used on an endpoint that returns a page, the dependency injection ends up redirecting to the appropriate login page.  which this may also do...

bnmnetp (changes requested):
Just one minor change...

bases/rsptx/assignment_server_api/assignment_builder/src/renderers/assignment.jsx
morozov-av: Of course! This code is not used in the new version of our pages, only for the old ones. I just uncommented it :)
We will remove it along with the removal of all the old pages for the Assignment Builder.

bases/rsptx/assignment_server_api/routers/instructor.py
morozov-av: At the moment, it's hard for me to either agree or disagree with you :) I just noticed that we duplicate the code for this check quite often, and I thought it would be interesting to move it into a separate decorator. I'd be happy to improve it, but let's not do it as part of this PR :)

components/rsptx/db/models.py
```diff
@@ -1069,3 +1069,9 @@ class Web2pySessionRunestone(Base, IdMixin):
     modified_datetime = Column(DateTime)
     unique_key = Column(String(64))
     session_data = Column(LargeBinary)
+
+class UpdateAssignmentExercisesPayload(BaseModel):
```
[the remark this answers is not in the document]
morozov-av: Thanks! Fixed

bnmnetp (approved):

PR #620, opened 2025-01-07
Bugfix: guide link for LTI in admin panel

ascholerChemeketa:
Fixes the link from the admin panel to the LTI section in the guide


bnmnetp:
Is there something I can do in terms of build configuration to keep PreTeXt from arbitrarily changing the output file name?

ascholerChemeketa:
I think the cause/fix is...
The book is building with chapters as the chunk (HTML page). Those don't have xml:id attributes. So they get numbers based on their position within the book, which does have an XML:id.

Try adding XML:id's to the chapters.

That will of course break my "fixed" link. But will be better in the long run.

bnmnetp:
OK, I didn't realize how inconsistent I was on adding `xml:ids` to chapters.  The link should now be at:  `https://guide.runestone.academy/managing-your-course.html#lti_integration`

bnmnetp:
I fixed this with the "new new" urls.

PR #653, opened 2025-02-03
Allow livecodes to read updated contents of editable datafiles

ascholerChemeketa:
Currently livecodes ignore any changes to editable datafiles.

`el.textContent` only returns the initial contents of the text area. `el.value` is required to get to the current contents.


bnmnetp (approved):
Thanks for the fix.

Issue #661, opened 2025-02-27
Codemirror related accessibility issues


oscarlevin:
Discussion on -dev: https://groups.google.com/g/pretext-dev/c/rP5JQcQH7_8

While trying to resolve the remaining accessibility errors in my book, I ran across a few that PreTeXt and Runestone should be handling but aren't.

In active code (and other code-highlighted contexts), the css appears to be using css from "codemirror-dark.less" (line 8) even when not in dark mode.  As such, the blue color for keywords does not have a strong enough contrast to pass tests. See for example https://discrete.openmathbooks.org/dmoi4/sec_seq_intro.html#subsec-seq-python

Again in active code, and probably elsewhere, we set a language attribute in a div as `lang="python"`.  WAVE complains that the language is invalid.  It appears that changing the attribute to `language="python"` works here.  But of course this needs to be set up to work with Runestone correctly.

The slider in active code contains an empty link: `<a class="ui-slider-handle ui-state-default ui-corner-all" href="#" style="left: 0%;"></a>` which causes an error.  Why should this be a link?

The textarea element of the active code is being flagged for missing a form label.  This can probably be fixed with some aria-label element?  The html it issue is: 
```<textarea autocorrect="off" autocapitalize="off" spellcheck="false" style="position: absolute; bottom: -1em; padding: 0px; width: 1000px; height: 1em; min-height: 1em;" tabindex="0"></textarea>```

Note that in the related issue #656, there was an error with a `<label>` element wrapping the free response textbox.  Removing this made everything okay, but there is `aria-label="textarea"` as an attribute in the `<textarea>` element.  So I think that's what we need here.

Lest you think I'm just picking on active code, I get the same missing label error for sage cells.  

Here is the reference from WAVE for the missing label issue: 

> How to Fix It
> If a text label for a form control is visible, use the <label> element to associate it with its respective form control. If there is no visible label, either provide an associated label, add a descriptive title attribute to the form control, or reference the label(s) using aria-labelledby. Labels are not required for image, submit, reset, button, or hidden form controls.
> The Algorithm... in English
> An <input> (except types of image, submit, reset, button, or hidden), <select>, or <textarea> does not have a properly associated label. A properly associated label is:
> a non-hidden <label> element with a for attribute value that is equal to the id of a unique form control
> a <label> element that surrounds the form control, does not surround any other form controls, and does not reference another element with its for attribute
> a non-empty title attribute, or
> a non-empty aria-labelledby attribute

PR #662, opened 2025-02-28
add screen reader support for activecode

oscarlevin:
I *think* this is what needs to be done to partially address #661, but testing is not working for me.  I can ask about how to better set up the javascript testing environment better at drop-in.


oscarlevin:
Okay, I figured out the testing and this does indeed resolve the issue with missing label on the textarea.

I'm going to follow up with a pr to (hopefully) clarify the documentation.

PR #664, opened 2025-02-28
darken pretext keyword color to fix contrast error

oscarlevin:
This addresses the contrast errors for keywords in activecode in light mode.

A PR to pretext to have their keywords match this color will follow.

PR #665, opened 2025-02-28
use data-lang attribute instead of lang for createEditor function

oscarlevin:
This addresses the penultimate issue related to active code accessibility: the use of `@lang` in a outerDiv to help determine the value of the programming langue for the textarea.  Switched to `@data-lang` and tested against the sample book.

### DIFF
--- a/bases/rsptx/interactives/runestone/activecode/js/activecode.js
+++ b/bases/rsptx/interactives/runestone/activecode/js/activecode.js
@@ -221,14 +221,13 @@ export class ActiveCode extends RunestoneBase {
         codeDiv.classList.add("ac_code_div");
         codeDiv.setAttribute("aria-label", "CodeMirror Editor");
         this.codeDiv = codeDiv;
-        this.outerDiv.lang = this.language;
         this.origElem.replaceWith(this.outerDiv);
         if (linkdiv.id !== this.divid) {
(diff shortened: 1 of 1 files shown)


bnmnetp (commented):
Looking at this now, I am wondering whether we even need to set the attribute on outerDiv !??  It does not appear that we reference it elsewhere.    It seems to me that we could just set `edmode` to  this.language.  Or not even have an edmode variable and simply refer to this.language.  I kind of take it back... we do reassign edmode appropriately but I don't think we need a data-lang attribute on outerDiv.

oscarlevin:
That's a great point.  It seems to be working with that outerDiv completely removed; it's only purpose was to hold an attribute we would immediately read for the purpose of setting another variable.  Maybe there were more uses previously.

Force pushed that change, plus a fix to add an aria-label to the slider's `<a>` element.  This fixes all the accessibility issues from #661 (once the contrast PR, #664  is merged as well).

This was a fun puzzle hunt.

bnmnetp:
I don't think we can eliminate this.outerDiv, it is used in many other places in activecode.js

oscarlevin:
Gotcha.  Now the outerDiv is still present, just without any attribute, and edmode is set to `this.language` directly.

PR #667, opened 2025-03-10
Better result formatting for cpp doctest unit tests

ascholerChemeketa:
Improvements to the parsing/output of doctest unit tests for cpp code.


ascholerChemeketa:
Not what I am seeing:
image

I was getting some undefined errors BEFORE this PR -w as one of the motivations. Are you sure it is running the new code?

PR #668, opened 2025-03-11
Grading page due date DST fix

ascholerChemeketa:
Issue raised by Debbie in Discord:
https://discord.com/channels/1013815439161315348/1013815440000159847/1348362626748387338

Grading page needs to use assignment date to call `getTimezoneOffest` and not `now`


bnmnetp (commented):
I'm not seeing how this would result in anything different than the current behavior.  The duedate is stored in the database sans any timezone information but the timezone info gets added by the Date constructor.  in either case.  

So, although the creation of the now variable is redundant I don't see how it will result in anything different.

bnmnetp:
Ahhh, I see the Date constructor is DST aware and will do the right offset for the given date.

bnmnetp:
fixes #669

PR #672, opened 2025-03-23
Update dragndrop.js

geoff-cox:
Adds the method `adjustDragDropWidths` to dynamically set widths (within constraints). Integrates calls to this method after rendering, resetting, and dragging/dropping items.

### DIFF
--- a/bases/rsptx/interactives/runestone/dragndrop/css/dragndrop.css
+++ b/bases/rsptx/interactives/runestone/dragndrop/css/dragndrop.css
@@ -9,8 +9,6 @@
     font-family: monospace;
     font-size: 120%;
     padding-bottom: 10px;
-    min-width: 48%;
-    max-width: 48%;
     border: 1px solid var(--componentBorderColor, #efefff);
     box-sizing: border-box;
 }
--- a/bases/rsptx/interactives/runestone/dragndrop/js/dragndrop.js
+++ b/bases/rsptx/interactives/runestone/dragndrop/js/dragndrop.js
@@ -148,7 +148,10 @@ export default class DragNDrop extends RunestoneBase {
         }
         self = this;
         this.ivp = this.isValidPremise.bind(this);
-        self.queueMathJax(self.containerDiv);
+        // Ensure MathJax has completed before adjusting the zone widths
+        self.queueMathJax(self.containerDiv).then(() => {
+            self.adjustDragDropWidths();
+        });
     }
 
     finishSettingUp() {
@@ -166,7 +169,6 @@ export default class DragNDrop extends RunestoneBase {
             this.dragDropWrapDiv.style.minHeight =
                 this.minheight.toString() + "px";
         }
-        this.adjustDragDropWidths();
     }
     addDragDivListeners() {
         let self = this;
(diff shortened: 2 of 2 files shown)


geoff-cox:
I accidentally sent this as I was reviewing it and before testing, but maybe this is a good way to discuss what else it may need. I believe, the dragndrop.css needs the min-width and max-widths removed. Is there anything else you can think of?

geoff-cox:
Here is the motivation:

This PR aims to accommodate wider a dropzone by adjusting the widths of the drag zone and drop zone panels based on their content. Currently, the component creates drag and drop panels of equal width. However, wide mathematical content in the drop zone spills outside the panel and can utilize the empty space in the drag zone.

geoff-cox:
Here is a screen of the use case.

matching

bnmnetp:
Thanks for the PR.  This looks like a good update.

My quick take on this is that you will want to ensue that MathJax has done all of its rendering before you start measuring widths.  If I recall correctly the call to `queueMathJax` returns a promise that is resolved when MJ is done.

If you don't then the results will be unpredictable.

geoff-cox:
Thanks for the feedback. If I am using `queueMathJax` correctly, the latest changes should check that MathJax has been rendered before calculating the zone widths.

geoff-cox:
I want to work on more interactive questions but am new to JS and web development. I would also like to create a minimalistic development environment to test and debug quickly. Is it reasonable/possible to make such an environment, or do I need to emulate a running runestone server to ensure accurate code implementations?

bnmnetp:
Thanks, I would welcome future contributions!

There are some more things to clean up on this one that you would see immediately if you had an environment set up.  So I'll try to give you some steps to follow that do not involve setting up a complete runestone server.

0. You already have the rs repo cloned.
1. Make sure you have poetry installed and these [prerequisites](https://runestone-monorepo.readthedocs.io/en/latest/poetry.html) 
2. From the root of the rs folder hierarchy run `poetry install --with=dev` and then `poetry shell`  This gets you in a python virtual environment with all the needed bits installed.
3. You should clone [pretext](https://github.com/PreTeXtBook/pretext) so that you have a copy of the sample book. Its under examples in the pretext repo.
4. You need to have node/npm installed a fairly recent version > 22 if I recall.
5. You will need to edit the project.ptx file for the sample book so the html target looks like this:
```
    <target name="html"
      format="html"
      source="sample-book.xml"
      publication="publication-noparts.xml" >
      <stringparams debug.rs.dev="yes" />
    </target>
```
Its the stringparam addition that is critical.

6. Now you can go ahead and build `pretext build html` but before you view you need to get your new javascript into place so its loaded with the page.
7. from the `.../rs/projects/interactives` folder run `./build.py --dev --static /path/to/pretext/examples/sample-book/output/html/_static`
8. Now you can go ahead and run `pretext view` in the sample book folder.

As you are developing and refining this PR you only need to repeat step 7 and 8 when you make changes to the javascript.

A lot of this is documented [here](https://runestone-monorepo.readthedocs.io/en/latest/javascript_feature.html) But could maybe be more clear.


OK, here are some screenshots of how the matching problems look with your PR applied to my dev environment.

Screenshot 2025-03-24 at 8 01 09 AM



Screenshot 2025-03-24 at 8 01 54 AM

After I click the reset button number 2 looks like this:

Screenshot 2025-03-24 at 8 03 00 AM

I don't think these will be too hard to fix up with a bit more css.

Happy to answer questions.  Hit me up in the discord or during drop in if you need any further help getting your environment set up.

PR #674, opened 2025-03-25
Flexible drag and drop widths

geoff-cox:
Adds the method adjustDragDropWidths to set widths (within constraints) dynamically. Integrates calls to this method after rendering, resetting, and dragging/dropping items. It also addresses the minimum height of the zones that do not reset after the reset button is pressed.

### DIFF
--- a/bases/rsptx/interactives/runestone/dragndrop/css/dragndrop.css
+++ b/bases/rsptx/interactives/runestone/dragndrop/css/dragndrop.css
@@ -9,6 +9,7 @@
     font-family: monospace;
     font-size: 120%;
     padding-bottom: 10px;
+    min-width: 28%;
     border: 1px solid var(--componentBorderColor, #efefff);
     box-sizing: border-box;
 }
--- a/bases/rsptx/interactives/runestone/dragndrop/css/dragndrop.css
+++ b/bases/rsptx/interactives/runestone/dragndrop/css/dragndrop.css
@@ -1,5 +1,6 @@
 .rsdraggable {
-    margin-left: 2%;
+    margin-left: 1%;
+    margin-right: 1%;
     margin-top: 5px;
     display:inline;
     float:left;
--- a/bases/rsptx/interactives/runestone/dragndrop/css/dragndrop.css
+++ b/bases/rsptx/interactives/runestone/dragndrop/css/dragndrop.css
@@ -9,8 +9,6 @@
     font-family: monospace;
     font-size: 120%;
     padding-bottom: 10px;
-    min-width: 48%;
-    max-width: 48%;
     border: 1px solid var(--componentBorderColor, #efefff);
     box-sizing: border-box;
 }
--- a/bases/rsptx/interactives/runestone/dragndrop/js/dragndrop.js
+++ b/bases/rsptx/interactives/runestone/dragndrop/js/dragndrop.js
@@ -345,12 +345,17 @@ export default class DragNDrop extends RunestoneBase {
                     // Make sure element isn't already there--prevents erros w/appending child
                     ev.target.appendChild(draggedSpan);
                 }
+                this.queueMathJax(this.containerDiv).then(() => {
+                    this.adjustDragDropWidths();
+                });
             }.bind(this)
         );
     }
 
     adjustDragDropWidths() {
-        // Ensure MathJax has completed when calling this method
+        // Temporarily minimize the dragzone width to the content
+        this.draggableDiv.style.width = "fit-content";
+        
         const dragzoneWidth = this.draggableDiv.offsetWidth;
         const totalWidth = this.dragDropWrapDiv.offsetWidth;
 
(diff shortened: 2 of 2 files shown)


geoff-cox:
Thanks for the feedback and guidance! I was able to set up the testing environment and refine this PR. This update seems promising.

bnmnetp:
Good, but I don't like this;
Screenshot 2025-03-25 at 6 22 32 PM
I think some minimum width would be good for both sides.

geoff-cox:
I agree that looks bad, but that is not what I see when testing. Here is what this problem looks like on my end.
git-pr-screen

bnmnetp:
I'm seeing what I'm seeing in both Firefox and Chrome.  Have you done a hard refresh of the page?  It looks to me like the min-width rule is still in effect from the css.  I've learned over time that during development you have to hard refresh the page or you can get confusing/unexpected results.

geoff-cox:
Yeah, I just tried a hard refresh on both Chrome and Firefox. Both of them show the empty dragzone with a minimum width of 28%

git-pr-screen2

bnmnetp:
If I reset the problem and drag everything out then it stays, but if I then refresh the page it goes to zero width as in the screen shot.

geoff-cox:
Perhaps I don't see this because I work outside a Runestone server. For me, a page refresh resets the problem to its initial state. 

I just pushed a min-width of 28% to the CSS. Does that fix it?

bnmnetp:
That would certainly explain things.  I'm a little surprised it didn't save your answers and restore them from localStorage.  Something to look into another day.

bnmnetp:
OK, that looks good.  This is what I see for the first problem when I drag things out.  In the current implementation the whole bordered blue box grows so that the drops don't extend below.

Screenshot 2025-03-26 at 3 37 59 PM

Finally I think we should fix the styling on the div containing the buttons to ensure they are always below the drag and drop zones and centered.

geoff-cox:
Okay, this is strange because everything looks perfect on my end, and I can't recreate what you are seeing. However, after looking at your screenshot, I see that our themes are slightly off. Could we be testing these on different versions of pretext CSS? Here is what I see for the same problem.

git-pr-screen3

bnmnetp:
Ahhhh, I think you hit on it.   @ascholerChemeketa I am using the "salem" theme on my end.

@geoff-cox You can edit the publication-runestone-academy.xml file so that in the html section the css element just reads:

```
<css theme="salem"/>
```
You will need to rebuild.  I think we want to make sure this works well with both "salem" and "denver" as I see those as the themes for the future.

That said, I have the book on Runestone Academy built with the salem theme and the blue container expands appropriately there.  So there is still a bit of mystery.  I'll check my local non-runestone version so see how it plays with your PR as well.

bnmnetp:
Ugh!!! I don't think it is an issue with Salem.

Strange issue,  But for me I had one answer of the drag and drop pre-dropped.  So it was restoring the half finished problem, ONLY then does the drop area grow outside of the blue container.   If I reset the problem and start over everything works as it should.

bnmnetp:
For standalone builds `publication-noparts.xml` is the correct file.

ascholerChemeketa:
Looks like some timing issues around rendering of the components and the attempt to resize them.

You may want to look into https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame. Changes to HTML from JS don't happen immediately, so you often have to use `requestAnimationFrame` or something similar to say "then run this function right before the next repaint".

geoff-cox:
I just updated the PR.

- It should now account for local storage. 
- It also dynamically adjusts the widths as items are dropped, eliminating the subtle overflow issue in the "Match each subspace" problem in the sample-book. It only shrinks if the largest item is dropped, so if there is more than one wide item, it will still spill over until there is room from the drag zone.
- I noticed that the minimum height does not reset after checking the answer and hitting the reset button. I believe this PR also addresses that issue by recomputing the zone height at the end of the reset code.

Perhaps I should have broken these into multiple PRs, but I am still learning the ropes. Please let me know if you have any feedback.

bnmnetp:
Thanks @geoff-cox one PR is good.  If it gets to be a lot of files or if it addresses a bunch of unrelated things then I would prefer to have things broken into multiple PRs.

I think @ascholerChemeketa might be on to something...  If I reset the problem and answer it things look good as below:
Screenshot 2025-03-27 at 12 17 30 PM

But once I refresh the page then they don't look so good.
Screenshot 2025-03-27 at 12 21 26 PM

My guess is this is related to restoring the answers.   One way to know that all that has happened is `this.component_ready_promise` it is resolved only after the data has been fetched.

geoff-cox:
Unfortunately, I can't recreate that issue, even when I switch to the Salem theme. My drop zone panel uses the entire width.

I cloned both RS and Pretext yesterday; other than the PR changes, everything is identical. I tested this on Chrome and Firefox.

If you are on Discord tomorrow, perhaps I'll drop in, and we can bounce some ideas around.

bnmnetp:
OK, I think I have a good fix.  I added a class to the div containing those two buttons and then added a couple of lines of css that give me good consistent behavior of those buttons.  I'm kind of surprised that was not already in place!

bnmnetp:
I'll fiddle a bit more and merge later today, so it can be live tomorrow.

geoff-cox:
Great, thanks! The PROTEUS group will be happy as we are currently working on matching problems. That was part of my motivation for working on this.

PR #675, opened 2025-03-31
Activecode locked regions improvements

ascholerChemeketa:
Two cleanups. 

First  commit changes max & min size on code mirrors. Sets a min to prevent resizing so small you can no longer grab resize handle. Max is set to always fit on screen with run button. Goal is that for long codemirrors, users scroll the code mirror and never have to scroll window while editing.

Second commit refactors the locked code decoration. Apparently inserting the gutter markers is rather inefficient. On exercise pages with 10+ exercises each with 20ish lines of locked code I was seeing 15+ second load times for code mirrors. This reduces the number of lock markers drawn by showing a line to indicate locked region (which I think looks better anyway). It also avoids drawing the lock markers in the initial editor creation - they snap in after a short delay. Finally, it adds subtly different background to locked lines to make it clearer where students are supposed to work.

image

### DIFF
--- a/bases/rsptx/interactives/runestone/activecode/js/activecode.js
+++ b/bases/rsptx/interactives/runestone/activecode/js/activecode.js
@@ -197,9 +197,12 @@ export class ActiveCode extends RunestoneBase {
             this.codeCoachList.push(new PyflakesCoach());
         }
 
+        // Why is this necessary???
         setTimeout(
             function () {
                 this.editor.refresh();
+                // need to regen locked decoration
+                this.setLockedRegions();
             }.bind(this),
             1000
         );
@@ -246,7 +249,7 @@ export class ActiveCode extends RunestoneBase {
         let gutterList = [];
         if (this.visiblePrefixEnd || this.visibleSuffixLength) {
             gutterList = [
-                { className: "CodeMirror-lock-markers", style: "width: 6px" },
+                { className: "CodeMirror-lock-markers", style: "width: 16px" },
             ];
         }
         var editor = CodeMirror(codeDiv, {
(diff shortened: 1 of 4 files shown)


ascholerChemeketa:
Converting to draft while I play with some related code...

ascholerChemeketa:
Did some more work on the initial commits after playing with saving only the user edited code (not the locked visible prefix/suffix).

Added another commit that modifies the saving/history to only work with user edited code.

ascholerChemeketa:
Think I found a bug in locked region filtering... back to draft

ascholerChemeketa:
OK, fixed an issue related to the 3rd commit (not saving locked portions of the edited code).

Switched back to "ready for review"

bases/rsptx/interactives/runestone/activecode/js/activecode.js
```diff
@@ -197,9 +200,12 @@ export class ActiveCode extends RunestoneBase {
             this.codeCoachList.push(new PyflakesCoach());
         }
 
+        // Why is this necessary???
```
bnmnetp, line 203: It was added in 2020.... I don't remember why, there was probably some kind of timing issue that this fixed.  May or may not even exist anymore.  Did you try removing it?

bnmnetp (commented):
This looks fine to me.

bnmnetp (approved):

bases/rsptx/interactives/runestone/activecode/js/activecode.js
ascholerChemeketa: I tried turning it off and didn't notice any issues. But I wasn't confident there isn't something out there relying on it.

PR #676, opened 2025-03-31
Install and use gcc13 in Jobe server

ascholerChemeketa:
GCC 13+ is required for using some of the nice C++20 features - in particular `std::format`.

This install gcc & g++ v13 and sets them as the default when invoking `gcc` or `g++`

### DIFF
--- a/projects/jobe/Dockerfile
+++ b/projects/jobe/Dockerfile
@@ -86,6 +86,18 @@ RUN ln -snf /usr/share/zoneinfo/"$TZ" /etc/localtime && \
     apt-get -y clean && \
     rm -rf /var/lib/apt/lists/*
 
+# Install and use more recent gcc and g++
+RUN apt-get update && \
+    apt-get --no-install-recommends install -yq \
+        software-properties-common \
+        gnupg-agent
+RUN add-apt-repository -y ppa:ubuntu-toolchain-r/test
+RUN apt-get update && \
+    apt-get install -y gcc-13 && \
+    apt-get install -y g++-13 && \
+    update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-13 60 && \
+    update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-13 60
+
 # Fix catch.hpp (use the one from catch2)
 RUN rm -f /usr/include/catch.hpp && ln -s /usr/include/catch2/catch.hpp /usr/include/catch.hpp
 


pearcej:
This sounds promising if it will work for Runestone more broadly. Perhaps we need a wider conversation iso n philosophy. I believe that with a language as old as C++ and with some of our alumni likely working with legacy code, students need to see the best and most essential of both older and newer C++.

bnmnetp:
gcc-13 was released in April of 2023, so I don't think we are on the bleeding edge of anything here, and I feel like we should try to keep relatively recent.  Although Java is far behind, I think that has to be that way due to the College Board and the APCSA curriculum.

Do we know of any deprecations or backward incompatibilities that might break current code?

ascholerChemeketa:
I don't know of any backwards compatibility issues. Books can set `-std=c++XX` as they please, so if someone wants to enforce ancient features they can opt for `-std=c++03`.

I interpreted Jan's comment as more about book updates/conversions. In the case of a single author that isn't a big deal. But for anything that is a semi-shared project there might need to a discussion about when to use fancy new tools and when to teach older methods.

bnmnetp:
OK, I'll let the C++ brain trust take the lead on that conversation.  I would imagine that if you and Jan and Jason are in alignment then I'm happy.

Tagging @wrigjl

wrigjl:
I don't think anything will break in thinkcpp or cppds, and anything that does can be easily fixed.

wrigjl:
This got me thinking though, there should be a way of extracting the source and ensuring that it compiles. It might make a good project for an undergrad/TA to:
1. parse the xml (and/or rst)
2. find source code
3. try to compile it
4. add a blacklist/whitelist for ignoring code snippets that aren't whole programs (maybe by xml:id)

ascholerChemeketa:
We're getting increasingly off topic, but expanding on Jason's suggestion... 

I've found it helpful to factor out code samples into separate actual code files that then get included in pretext with something like:

```
<program label="input-output_format-spacing" interactive="activecode">
<xi:include href="../../programs/input-output/format-spacing.cpp" parse="text"/>
</program>
```

It eliminates the need for CDATA or lots of `&lt;`s. It also makes it much easier to do things like running a formatting tool or compiler on them.

So part of the student job could be to start by refactoring all the samples that are intended to be runnable programs to point to external files.

pearcej:
I am fine with this, but since C++ was standardized I believe that many C++ compilers have become more stringent about non-standard code, so, yes, of course, it is about textbooks, but also about not breaking them without notice, especially not in the final months of a term. I would expect all textbooks are very largely using standards complete code, but there might be sections that don't for educational reasons, and I would hate to see them break if an author was not a Discord user.

bnmnetp:
So, maybe best to put a pin in this as far as merging goes until things have wrapped up for the year?  Merging towards the end of may should be pretty safe.

pearcej:
Even though I am likely being overly cautions, I think the end of May is safer.

In the meantime, if she doesn't respond here, @bnmnetp can you talk to @barbarer in your regular meeting to ensure this is fine?

ascholerChemeketa:
Yes, fine to wait.

99.99% sure the current version of GCC on Jobe defaults to `-std=gnu++14`. So everyone already is using standardized C++. GCC13 makes the default `-std=gnu++17` .

Ideally every author sets compiler options for their book. In PreTeXt you can add something like this to your docinfo and make sure you are getting the version of the standards you want:
```
    <programs language="cpp"
              compiler-args=" -Wall -Wextra -std=c++20 -fmodules-ts -Wno-unused-variable -Wno-unused-parameter"
              timelimit="5000"/>
```

ascholerChemeketa:
After more testing of gcc13 vs 14 I strongly prefer moving to 14.

Doing so painlessly requires updating the Jobe server to run on Ubuntu 24.04 (from 22.04).

Updating to 24.04 would require updating openjdk from `openjdk-18-jdk` to 21 (or downgrading to 17) as 18 was not a LTS release of the openjdk and it does not exist in 24.04 v17 and v21 are both LTS releases. v21 has been available since 2023. The next LTS release (v25) is not due until this fall.

I have tested openjdk-21 against the Pretext Sample Book and against CSAwesome. I could not find any issues in CS Awesome and everything that currently works in PTXSB still works (a number of the java programs in it have existing issues).

I updated the PR to make these upgrades.

I could change it keep Ubuntu 22.04 and thus openjdk-18, but that would require building gcc from source which is a multi-hour process.

I'll ping authors on Discord re: jdk version. But it seems like it would make sense to move to an LTS release.

bnmnetp:
I agree, we don't want to let JOBE stagnate, and I think the new curriculum for CSA including changes to CSAwesome use feature from 21.

PR #694, opened 2025-04-15
Add math-mode support for horizontal Parsons problems

geoff-cox:
This patch extends the Runestone hparsons component to support math-mode horizontal Parsons problems.

### DIFF
--- a/bases/rsptx/interactives/runestone/hparsons/js/hparsons.js
+++ b/bases/rsptx/interactives/runestone/hparsons/js/hparsons.js
@@ -25,6 +25,10 @@ export default class HParsons extends RunestoneBase {
         this.randomize = $(orig).data("randomize") ? true : false;
         this.isBlockGrading = $(orig).data("blockanswer") ? true : false;
         this.language = $(orig).data("language");
+        // Detect math mode
+        if (this.language === undefined && orig.textContent.includes('span class="process-math"')) {
+            this.language = "math";
+        }
         if (this.isBlockGrading) {
             this.blockAnswer = $(orig).data("blockanswer").split(" ");
         }
@@ -46,6 +50,9 @@ export default class HParsons extends RunestoneBase {
         this.outerDiv = null;
         this.controlDiv = null;
         this.processContent(this.code);
+
+        this.microParsonToRaw = new Map();
+        this.simulatedSolution = [];
 
         // Change to factory when more execution based feedback is included
         if (this.isBlockGrading) {
@@ -83,10 +90,10 @@ export default class HParsons extends RunestoneBase {
         this.unittest = this.processSingleContent(code, "--unittest--");
     }
 
-    processSingleContent(code, delimitier) {
-        let index = code.indexOf(delimitier);
+    processSingleContent(code, delimiter) {
+        let index = code.indexOf(delimiter);
         if (index > -1) {
-            let content = code.substring(index + delimitier.length);
+            let content = code.substring(index + delimiter.length);
             let endIndex = content.indexOf("\n--");
             content =
                 endIndex > -1 ? content.substring(0, endIndex + 1) : content;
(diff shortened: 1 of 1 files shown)


geoff-cox:
Here is one of the test cases I used in Pretext:

```
<exercise label="horizontal-parson-math-simple" language="natural"><title>Parsons Problem, Math</title>
    <statement>
        <p>
            Form the SQL statement by rearranging the four blocks.
        </p>
    </statement>
    <blocks layout="horizontal" reuse="no" randomize="no">
        <block order="4"><m>\alpha</m></block>
        <block order="2"><m>\beta</m></block>
        <block order="1"><m>\displaystyle\sum_{i=1}^{\infty}</m></block>
        <block order="3"><m>x^2</m></block>
        <block order="5" correct="no"><m>\displaystyle\int x^2</m></block>
    </blocks>
</exercise>
```

For quick reference, here is the correct order
image

bases/rsptx/interactives/runestone/hparsons/js/hparsons.js
```diff
@@ -152,13 +160,64 @@ export default class HParsons extends RunestoneBase {
...
+
+    renderMathInBlocks() {
+        if (this.language !== "math") return;
+        setTimeout(() => {
+            const blocks = document.querySelectorAll(`#${this.divid}-container .parsons-block`);
+            blocks.forEach(block => {
+                block.innerHTML = this.decodeHTMLEntities(block.innerHTML);
+            });
+
+            if (window.MathJax && MathJax.typesetPromise) {
```
bnmnetp, line 186: Thanks for this addition!

You should have a look at parsons.js and others for how we use the this.queueMathJax(specific item).

MathJax can create a big set of timing problems and this allows us to take control.  Also trying to re-typeset the whole page can be a performance issue.

bases/rsptx/interactives/runestone/hparsons/js/hparsons.js
```diff
@@ -152,13 +160,64 @@ export default class HParsons extends RunestoneBase {
...
+                MathJax.typesetPromise().then(() => this.simulateSolution());
+            }
+        }, 0);
+    }
+
+    simulateSolution() {
```
bnmnetp, line 192: Maybe some comments about what the `simulateSolution` function is all about?  Its not clear to me.. Is this for scoring?  I would have guessed that block scoring would be fine?  Maybe not for some expression.

geoff-cox: @bnmnetp  
The Micro-parsons code was formatting the math blocks with unpredictable Ascii characters, so I simulated the correct solution in the answer area and asked Micro-parsons to output what a correct solution would look like from their end. I needed this because a correct answer is determined by comparing the text on the blocks. I was hoping there would have been an ID on the blocks, but I didn't see any, so this was my workaround. I am sure there is a better way. Let me know if you have any ideas.

I can add more details to the PR tomorrow.

bnmnetp: Thanks for clarifying!  I assume all of the `\` ing of characters was the cause of the weird text.  

I does not look like the html representation lends itself easily to adding ids.... But it seems like it should.  maybe worth investigating with Rob, and looking at a better html (or json) representation to use in the javascript.

I'm ok with your solution for now, but thinking long term.

geoff-cox: I think adding IDs could streamline future improvements and make grading in blockGrader.js more straightforward. 

I designed this thinking that it would be a good way to construct formulas and equations term-by-term. Unfortunately, it doesn't currently support multiple correct orderings, which would be essential for commutative operations like addition and multiplication. So getting Rob involved in coordinating how multiple orderings can be communicated to Runestone would be good.

geoff-cox: Also, display math tags (<me>) are not currently supported with this PR, but <m>\displaystyle * </m> is good to go.

PR #697, opened 2025-04-23
RUN-42

DanyaMarchenko:
Moved all exercise descriptions from frontend to backend. Used them on the exercise creation page


bases/rsptx/assignment_server_api/assignment_builder/src/hooks/useExerciseTypes.ts
```diff
@@ -10,6 +10,6 @@ export const useExerciseTypes = (): ExerciseTypeConfig[] => {
     ...option,
     color: colorSchemes[index],
     tag: option.value,
-    description: exerciseDescriptions[option.value]
+    description: option.description
```
morozov-av, line 13: Since we're already doing ...option above, this line is absolutely unnecessary. Please review [this documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax).

morozov-av (approved):

PR #699, opened 2025-04-23
Program add-files and compile-with

ascholerChemeketa:
Adds:
* separate acid and filename fields in `source_code` table
* processing of `<source>` elements in RS manifest file to record initial source code of programs
* support for `add-files` and `compile-with`
* looking up initial source code from `source_code` table if needed include is not on page and there is no student answer

This goes with: https://github.com/PreTeXtBook/pretext/pull/2472
This code should work fine with existing PreTeXt books, but new features depend on that PR.

Testing suggestion:
I tested this on the unmodified sample book and confirmed the existing data files all still work fine. The flowers datafile example does not work, but that is due to an existing bug (missing datafile) and is fixed in the updates to sample-book.
Then, build updated sample book to confirm things still work and new samples work.


ascholerChemeketa:
@bnmnetp Found and fixed the datafile issue that affected RST datafiles. Was missing a spot where I needed to look at the id for the filename instead of data-filename attribute.

Force pushed a change that only modified the last commit.

I built a copy of csawesome as well and confirmed that all of its data file examples work.

PR #705, opened 2025-04-28
Bugfix for RST datafiles lookups

ascholerChemeketa:
Bug fix for the recent PR.

Debbie Yuster reported a bug with cross book lookup of datafiles. I think the issue is datafiles that are in the DB without a filename in the source_code table. The migration from previous PR took care of old builds, but I think rebuilds were still missing filenames.

This has fixes at both ends:
*Makes sure the RST datafile build script adds filenames to the source_code table
*Modifies the crud.py `fetch_source_code` function to, when a datafile is requested by filename, match the filename against either the filename OR the acid.

Should not need a book rebuild for `fetch_source_code` to fix issue Debbie reported. But first fix will make sure everything is in DB correctly on build.

### DIFF
--- a/bases/rsptx/book_server_api/routers/rslogging.py
+++ b/bases/rsptx/book_server_api/routers/rslogging.py
@@ -616,7 +616,7 @@ async def get_source_code(
     """
     course = await fetch_course(course_id)
 
-    rslogger.debug(f"get_source_code: -{acid}-{filename}-")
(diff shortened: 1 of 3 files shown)


bnmnetp:
Still was not working for Debbie's example... But I found it:  in `activecode.js` we have 
```
url: `/ns/logger/get_source_code?course_id=${eBookConfig.course}&fileName=${fileName}`,
```
but in the get_source_code endpoint the parameter is `filename`

I fixed it and will push out runestone 7.6.1

ascholerChemeketa:
Doh. Sorry. Thanks for the catch.

PR #709, opened 2025-04-30
Activecode - set large memorylimit if code tries to use fsanitize

ascholerChemeketa:
The -fsanitize compiler flag generates code that requires a large shadow block of memory which exceeds the Jobe  default limit. This detects attempts to use fsanitize and tells Jobe to set a very high memory limit.


bnmnetp (approved):
seems fine to me.

PR #713, opened 2025-05-07
RUN-37

DanyaMarchenko:
Added tooltips to the exercises table with exercise details

### DIFF
--- a/bases/rsptx/assignment_server_api/assignment_builder/src/components/routes/AssignmentBuilder/components/exercises/AssignmentExercisesList/AssignmentExercisesTable.tsx
+++ b/bases/rsptx/assignment_server_api/assignment_builder/src/components/routes/AssignmentBuilder/components/exercises/AssignmentExercisesList/AssignmentExercisesTable.tsx
@@ -112,7 +112,7 @@ export const AssignmentExercisesTable = ({
                 className="pi pi-info-circle"
                 id={`info-icon-${data.id}`}
                 style={{ cursor: "pointer" }}
-              ></i>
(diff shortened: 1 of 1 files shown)


bases/rsptx/assignment_server_api/assignment_builder/src/components/routes/AssignmentBuilder/components/exercises/AssignmentExercisesList/AssignmentExercisesTable.tsx
```diff
@@ -59,6 +60,17 @@ export const AssignmentExercisesTable = ({
       !!exercise.question_json
     );
   };
+  const getTooltipText = (data: Exercise) => {
+    return Object.entries({
+      Author: data.author,
+      Difficulty: data.difficulty,
+      Tags: data.tags,
+      Chapter: data.chapter
+    })
```
morozov-av, line 69: What if difficulty is set to 0? What will happen in that case?

morozov-av: Yes, you're right - at the moment we can't set difficulty to zero, but it's still a numeric value, so it would be better to account for that. We might change this behavior in the future, and in that case, it would result in a bug here where zero wouldn't be displayed in your tooltip. So please make sure to prevent that in advance. Everything else looks great - excellent work!

@bnmnetp please review the PR

PR #714, opened 2025-05-12
RUN-36/RUN-32

morozov-av:
image
image

I’ve added dnd/matching  exercise creation and generated the structure according to your example, but somehow it’s not working for me. Maybe I need to rebuild something? I’ve already rebuilt the assignment and runestone parts, but it’s still not working on my side locally


bnmnetp:
Any parse errors in the javascript console?

bnmnetp:
copy and past the script tag above here, and I'll try to help debug.

morozov-av:
Hi @bnmnetp please have a look at this fix https://github.com/RunestoneInteractive/rs/pull/714/commits/32e2f910eacd120fab5807ac3581d870678edae2

Maybe you just forgot to add it, or maybe I was supposed to do something differently - but after this fix, everything started working for me. It seemed like our new builder wasn’t properly connected to the React part on my side.

bnmnetp:
Good catch, I did for get that, but the matching examples always came after the drag and drop examples in testing so it worked.

PR #718, opened 2025-05-16
Codelens questions update

ascholerChemeketa:
Updates Codelen to handle questions with HTML controls. With each answered question a score update is sent with partial credit based on the number of questions answered during this session. 

Also allows for specifying starting step for the trace.

I tested this against a `Overview` RST build and it seems to work fine. Then tested against sample-book prior to any changes in it. Also seems fine. 

I *believe* it is fine to merge this prior to action on partner PR: https://github.com/PreTeXtBook/pretext/pull/2513

(It looks like question scoring was completely off for codelens based on a typo in `EVENT2TABLE`.)


bnmnetp:
Love it! Thanks for the update.

bnmnetp:
Now available in runestone 7.7.0 and beyond

PR #719, opened 2025-05-20
Activecode add bindings for ctrl-s and ctrl-/

ascholerChemeketa:
Adds ctrl-s (Save & Run) and ctrl-/ (toggle comment on selection) keybindings to activecode editors.

### DIFF
--- a/bases/rsptx/interactives/runestone/activecode/js/activecode-i18n.en.js
+++ b/bases/rsptx/interactives/runestone/activecode/js/activecode-i18n.en.js
@@ -111,6 +111,7 @@ $.i18n().load({
             "Compiling and Running your Code Now...",
         msg_activecode_server_comm_err: "Error communicating with the server.",
         msg_activecode_save_run: "Save & Run",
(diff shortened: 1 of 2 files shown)


bases/rsptx/interactives/runestone/activecode/js/activecode-i18n.en.js
```diff
@@ -111,6 +111,7 @@ $.i18n().load({
             "Compiling and Running your Code Now...",
         msg_activecode_server_comm_err: "Error communicating with the server.",
         msg_activecode_save_run: "Save & Run",
+        msg_activecode_run_button_title: "Save & Run (ctrl-s)",
```
bnmnetp, line 114: Is there another way to make the ctrl-s discoverable?  This seems like it is going to make the save & run button a lot wider.

ascholerChemeketa:
`title` == tooltip. You only see it on hover.

bnmnetp:
Missed that! Thanks.

ascholerChemeketa:
But yes, a model help on all components does seem like a good idea.

PR #721, opened 2025-05-21
use is not None test to prevent futurewarning

wrigjl:
when building a pretext book, I see this warning:

```
/Users/jason/rs/components/rsptx/build_tools/core.py:692: FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
  if el and "practice" in el.attrib:
```

The fix is easy enough, just use `is not None` test.


bnmnetp (approved):
Thanks

PR #723, opened 2025-05-21
Temporary patch: handle sync or async context in manifest_data_to_db

ascholerChemeketa:
Tested in process-manifest and build - both contexts successfully work.

I didn't add nest_asyncio to explicit dependencies in pyproject.toml files as we appear to be getting it indirectly and this isn't intended to stick around.

(Though the fact that we are getting it indirectly does mean that we are dependent on it... not sure if that makes your big fix more or less urgent.)


bnmnetp:
I think it might be coming in through as an ipython / jupyter dependency.  I think I remember reading that as I was searching for nest-asyncio

PR #727, opened 2025-05-25
RUN-43

DanyaMarchenko:
Fix wrong difficulty and points values in exercise creation. Update exercise creation page entries for difficulty and points. Fix plus icon hiding in assignment exercises table.


bases/rsptx/assignment_server_api/assignment_builder/src/components/routes/AssignmentBuilder/components/exercises/components/CreateExercise/shared/BaseExerciseSettingsContent.tsx
```diff
@@ -161,7 +161,7 @@ export const BaseExerciseSettingsContent = <T extends BaseExerciseSettings>({
               min={1}
               max={5}
               className="w-full"
-              onValueChange={(e) => updateSetting("difficulty", e.value !== null ? e.value : 3)}
+              onChange={(e) => updateSetting("difficulty", e.value !== null ? e.value : 1)}
```
morozov-av, line 164: Please avoid code duplication. create a separate function inside the component and use it in both places

bases/rsptx/assignment_server_api/assignment_builder/src/store/exercises/exercises.logic.api.ts
```diff
@@ -20,24 +20,13 @@ export const exercisesApi = createApi({
     createNewExercise: build.mutation<number, CreateExercisesPayload>({
       query: (body) => ({
         method: "POST",
-        url: "/assignment/instructor/new_question",
+        url: "/assignment/instructor/question_creation",
```
morozov-av, line 23: Would it be better to call the endpoint just POST /assignment/instructor/question ?

I assume the HTTP POST method already gives a clear idea of what we're doing here. What do you think, @bnmnetp ?

DanyaMarchenko: I agree that it's better. Renamed it in the new commit

bnmnetp:
Sorry for letting this linger.  I think the conflicts should be easy to resolve and then I will get this merged.

Issue #635, opened 2025-01-16
make aria-live regions for errors and stdout on activecode


bnmnetp:

jakeob28:
Hi,

I see this was marked as "done" on the Kanban, but I'm still seeing some aria tags missing on my local instance.

Testing the JavaScript activecode tool in section 1.10 of the overview book, it appears only stdout has `aria-live` set, and not the errors:

![Image](https://github.com/user-attachments/assets/70a76d57-d332-4cc3-bd01-f1f5db8a4abf)

Interestingly, testing activecode in C in section 1.15, there are two `errinfo` divs, one with `aria-live` set that is not used, and another without `aria-live` set that is actually populated with the error:

![Image](https://github.com/user-attachments/assets/50b9c805-44a5-47e1-9e96-60c2ca5a7de2)

This div without `aria-live` is used for both compilation errors and regular program output to stderr. I didn't find a scenario in which the first `errinfo` div is populated.

Finally, when I used the scratch activecode window for python, it seems that while stdout and errinfo had the correct `aria-live` tags, the code coach did not:

![Image](https://github.com/user-attachments/assets/3ca13be7-ef32-4643-ad2b-61993308873b)

Thanks! I'm happy to try and take a look at why this happens, although I haven't gotten a chance to look at the code yet.

bnmnetp:
I just dragged it to done this morning, thinking I had it all done.  But I guess more to do.  My focus had been on python since that is what is being used in the course at VT.

Any help you can give in tracking this down is great.  I'm no expert and just trying to do my best  to upgrade code that was written before all of these standards.

bnmnetp:
Also, I would have you focus more on examples from the pretext sample book (see PTXSB in the production site) The overview book was written in restructuredText and we have deprecated that in favor of PreTeXt which is much much better for accessibility.  The runestone components are the same in both however.

PR #737, opened 2025-06-05
Fix aria-live tags for activecode error/stdout

jakeob28:
This PR fixes some cases where `aria-live` tags were missing for activecode results. Notably:

- Codecoach outputs now have proper tags (bb3d3924f0144085ffcf743efcf6d82473f935b2)
- Javascript errors are properly tagged (5503d7ae5e26bc3f7359553eac883713a374d4fa)
- Languages that use JOBE have their errors properly tagged (a5b1ca96e11165e7786ad85cea5e2b25b093556d)
- SQL results are properly tagged (7e2db26b770cd74cf71d9414783e883299dacc27)

These issues were caused by the fact that originally the tags were only added in `activecode.js`, and not the other files that can generate their own error/output divs (see diff for details on where this happens).

This PR resolves RunestoneInteractive/rs#635.

### DIFF
--- a/bases/rsptx/interactives/runestone/activecode/js/activecode.js
+++ b/bases/rsptx/interactives/runestone/activecode/js/activecode.js
@@ -1003,6 +1003,7 @@ export class ActiveCode extends RunestoneBase {
 
         var coachDiv = document.createElement("div");
         coachDiv.classList.add("alert", "alert-warning", "codecoach");
+        coachDiv.setAttribute("aria-live", "polite");
         $(coachDiv).css("display", "none");
         let coachHead = coachDiv.appendChild(document.createElement("h3"));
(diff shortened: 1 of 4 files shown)


bnmnetp:
Thanks for this.  Did you check the results with a screen reader?  I had to use a setTimeout (see activecode.js line 1831) in order for the screen reader to pick up on the change after the container was made visible.  But maybe other screen readers are different I was testing with macOS and using Apple Voice Over.

jakeob28:
I just went ahead and tested with NVDA screenreader on windows and indeed saw that it missed errors without the delay. I added those in and now all the errors are reading out properly on my end.

bnmnetp:
Great!  Thanks for the assist.  I'll get this merged and released before the weekend rebuilds.  Then we can truly move this one into the done column.

PR #745, opened 2025-06-09
Fix activity label contrast

jakeob28:
When logged in as an instructor, currently portions of the activity labels currently have insufficient contrast per WCAG guidelines:
image

This PR fixes this by using italics instead of reduced opacity to indicate the activity identifier:
image

Let me know if you'd prefer a different method (other than italics) of indicating the assignment identifier. One alternative is just to use the parenthesis without any additional styling, which would look like:
image
or bolding the rest of the caption:
image


bnmnetp:
So we made them very light because many instructors find them distracting and unsightly.  I wonder if there is a way to satisfy WCAG guidelines while not drawing attention to them.

pearcej:
What about parentheses and right justification, so it is out of the top-left primary viewing area?
image

bnmnetp:
this is stale now and has been handled.

PR #746, opened 2025-06-10
Fix multiple choice question accessibility issue

jakeob28:
For multiple choice questions:
image
the HTML currently has a `<legend>` element with "choose one" or "choose multiple", and the actual question content is outside of this element:

```html
<div id="question1_1" class="runestone-component-ready">
  <p>What programming language does this site help you to learn?</p>
  <form id="question1_1_form" method="get" action="" onsubmit="return false;">
    <fieldset>
      <label>
        <p><input type="radio" name="group1" value="0" id="question1_1_opt_0">A. Python</p>
      </label>
      <br>
       ... (other answers elided for brevity)
      <legend>Choose one</legend>
    </fieldset>
    <button class="btn btn-success" name="do answer" type="button">Check Me</button>
  </form>
  <br>
  <div id="question1_1_feedback" aria-live="polite" role="status"></div>
  <p class="runestone_caption">Activity: 2.1.1 Multiple Choice</p>
</div>
```

This isn't all that useful to screenreaders, which expect the `<legend>` element to contain the actual question and will read it when one of the answer choices are focused for the first time. [See how MDN structures radio groups](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/radio) for reference.

This PR includes the entirety of the question inside the `<legend>` element. Visually there are no changes, but screenreaders now behave properly when tabbing over answer choices in my testing.

Here's an example of the new html:
```html
<div id="question1_1" class="runestone-component-ready">
  <form id="question1_1_form" method="get" action="" onsubmit="return false;">
    <fieldset>
      <label>
        <p><input type="radio" name="group1" value="0" id="question1_1_opt_0">A. Python</p>
      </label>
      <br>
       ... (other answers elided for brevity)
      <br>
      <legend>
        <p>What programming language does this site help you to learn?</p>
        <div class="instructions">Choose one</div>
      </legend>
    </fieldset>
    <button class="btn btn-success" name="do answer" type="button">Check Me</button>
  </form>
  <br>
  <div id="question1_1_feedback" aria-live="polite" role="status"></div>
  <p class="runestone_caption">Activity: 2.1.1 Multiple Choice</p>
</div>
```

### DIFF
--- a/bases/rsptx/interactives/runestone/mchoice/js/mchoice.js
+++ b/bases/rsptx/interactives/runestone/mchoice/js/mchoice.js
@@ -140,7 +140,10 @@ export default class MultipleChoice extends RunestoneBase {
 
     renderMCContainer() {
         this.containerDiv = document.createElement("div");
-        $(this.containerDiv).html(this.question);
+        this.questionDiv = document.createElement("div");
+        $(this.questionDiv).html(this.question);
+        this.questionDiv.id = this.divid + "_prompt";
+        this.containerDiv.appendChild(this.questionDiv);
         $(this.containerDiv).addClass(this.origElem.getAttribute("class"));
         this.containerDiv.id = this.divid;
     }
@@ -154,7 +157,9 @@ export default class MultipleChoice extends RunestoneBase {
             onsubmit: "return false;",
         });
         // Add fieldset and legend for accessibility
-        this.optsFieldSet = document.createElement("fieldset")
+        this.optsFieldSet = document.createElement("fieldset");
+        this.optsFieldSet.role = "radiogroup";
+        this.optsFieldSet.setAttribute("aria-labelledby", this.divid + "_prompt");
         this.optsForm.appendChild(this.optsFieldSet);
         // generate form options
         this.renderMCFormOpts();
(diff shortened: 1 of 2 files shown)


bnmnetp:
When I view problems written in PreTeXt with this change I get the following:

Screenshot 2025-06-11 at 4 22 50 PM

need to get rid of that `[object HTMLDivElement]` piece

jakeob28:
Upon further investigation, while I was able to get the HTML to be structured properly for both RST and PreTeXt, getting the styles to work for both types and all themes would've been a huge pain, since by putting the question inside the `<legend>` element I'm fighting the existing styles for `legend`. 

I ended up opting to revert the original change, and just add aria tags instead of changing the html structure. It's not ideal (since semantic html is generally preferred over aria tags), but it's still an acceptable solution and my screenreader behaves correctly on both rst and ptx books.

bases/rsptx/interactives/runestone/mchoice/js/mchoice.js
```diff
@@ -154,7 +157,9 @@ export default class MultipleChoice extends RunestoneBase {
             onsubmit: "return false;",
         });
         // Add fieldset and legend for accessibility
-        this.optsFieldSet = document.createElement("fieldset")
+        this.optsFieldSet = document.createElement("fieldset");
+        this.optsFieldSet.setAttribute("role", "radiogroup");
```
bnmnetp, line 161: Some multiple choice questions can have multiple answers in which case it is not a radio group but a bunch of check boxes.  Do you maybe want to check `this.multipleanswers` before setting the role to radiogroup?

bnmnetp: I see that they still look and work fine in both cases so maybe nothing to worry about.

PR #749, opened 2025-06-11
Fix focus on activecode modal close

jakeob28:
When the activecode modal is closed, focus is intended to return to the pencil icon (that was used to open the modal originally). There is currently code for this:
```js
let closeBtn = document.querySelector(".ac-modal-header button.close");
closeBtn.addEventListener("click", function () {
    let popUp = document.getElementById("ac_modal_" + divid);
    popUp.style.display = "none";
    document.querySelector("button.activecode-toggle").focus();
});
```
However, it errors since `document.querySelector("button.activecode-toggle")` fails to find any elements:
image
I'm guessing this is some sort of regression where the HTML was changed after this feature was implemented.

This PR fixes the selector/html so that the functionality works properly again.

### DIFF
--- a/bases/rsptx/interactives/runestone/activecode/js/acfactory.js
+++ b/bases/rsptx/interactives/runestone/activecode/js/acfactory.js
@@ -183,7 +183,7 @@ export default class ACFactory {
         closeBtn.addEventListener("click", function () {
             let popUp = document.getElementById("ac_modal_" + divid);
             popUp.style.display = "none";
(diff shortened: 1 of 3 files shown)


bnmnetp:
I think the change was made for the way PreTeXt builds its nav bar and it is a button.  Can you make it a button instead of an a for the RST too so it works for both?

jakeob28:
fix was pretty simple — just select on the class instead of including element type as well. Focus now works properly on my end for both rst & pretext with both the close button and escape key.

PR #752, opened 2025-06-12
Check that the basecourse/book has been built before creating a course from it

skiadas:
If a course is created via `rsmanage --addcourse` from a basecourse/book before that book has been built, then the needed attributes don't exist yet and are not copied over. This PR adds a check to prevent a user from accidentally creating a course too soon.


bnmnetp (approved):
Thanks

PR #773, opened 2025-06-20
RUN-38 Implement routing for assignment builder

morozov-av:


bnmnetp:
Thanks, The back button now works as expected.  Unfortunately the Create New Assignment button does not work with these changes applied.

bnmnetp (changes requested):
Bug - the Create New Assignment button is broken now.  Pressing it appears to do nothing and if you refresh the page it tries to load builder/create which is not found.

morozov-av:
Hmm, very strange, working for me.. I'll doublecheck @bnmnetp

morozov-av:
Could you please try to rebuild the project? @bnmnetp

morozov-av:
Found a bug, fixing @bnmnetp

morozov-av:
@bnmnetp Fixed, please have a look

PR #774, opened 2025-06-20
RUN-44 Add Section selector to settings

morozov-av:
This PR also includes changes to the routing, so the updates specific to the sub-section will only be visible after the routing is merged. Alternatively, you can go through the commits related to the sections separately.

@bnmnetp


morozov-av:
> It looks like I only need to merge this branch correct?

Yeah, correct. You asked me to make a separate PRs, but routing PR included the changes I needed

bnmnetp:
I get an error
```
base: basedir,F
```
that looks like a typo
Can you check?  if I remove the 'F' all is good

morozov-av:
> I get an error
> 
> ```
> base: basedir,F
> ```
> 
> that looks like a typo Can you check? if I remove the 'F' all is good

Arrrrrrrgh, yes, sorry. missclick before commit. fixed

PR #838, opened 2025-06-25
Fixes for readings

morozov-av:
Fix activity count being 0 in some cases. Set the number of required activities default to 80% of the activity count. Remove autograde column for selected readings

### DIFF
--- a/bases/rsptx/web2py_server/applications/runestone/controllers/admin.py
+++ b/bases/rsptx/web2py_server/applications/runestone/controllers/admin.py
@@ -2448,7 +2448,7 @@ def add__or_update_assignment_question():
         try:
             activities_required = int(request.vars.get("activities_required"))
             if activities_required == -1:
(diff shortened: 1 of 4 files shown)


bases/rsptx/web2py_server/applications/runestone/controllers/admin.py
```diff
@@ -2448,7 +2448,7 @@ def add__or_update_assignment_question():
         try:
             activities_required = int(request.vars.get("activities_required"))
             if activities_required == -1:
-                activities_required = max(int(activity_count * 0.8), 1)
+                activities_required = max(round(activity_count * 0.8), 1)
```
bnmnetp, line 2451: Oooh, this needs to go on my todo list, to port this to the assignment server.

bnmnetp: OK, I was not finding where this was used.  But I made a start anyway.  I'll keep it on a branch in case we need it.

PR #845, opened 2025-06-26
PI Text Chat Interface Redesign

ibrahimmoazzam:
This PR deals with the design overhaul of the text chat interface for Peer Instruction. 

It is currently a draft, and I will update the status once it's complete in the next few days.


bnmnetp:
I would prefer that the poetry.lock files for unrelated services be excluded from this PR, so it is just focused on the chat interface.

bnmnetp:
the conflict with the percent correct looks easy to resolve with my latest commit

PR #880, opened 2025-07-09
Issue 879. Fix creating timed and peer assignments

### LINKED ISSUE #879: Error when creating a new timed assignment
I was in the CSAwesome2 ebook and tried to create a new timed assignment.  When I clicked on the "Create" button a message flashed up and then disappeared with a fire icon "Error Creating Assignment".  I don't think error messages should disappear like this.

Image

DanyaMarchenko:
Fix #879


bnmnetp:
I should have marked that I was working on this.  These are the exact changes that I already committed.

See my comment on making the Toast persistent when we have an error.  Its important for instructors to be able to capture that when they report a problem.

PR #896, opened 2025-07-15
Issue 879. Toast enhancements

morozov-av:
Make the toast for error messages keep until clicked on. Add ability to dismiss any toast by clicking on it.


morozov-av:
@bnmnetp In my variant, there is no close button on the toast. Do you think we need to improve it and add the close button on the toasts?

PR #928, opened 2025-07-24
Mchoice enhancement

morozov-av:
## Mhoice exercise creation interface update
Add statement preview to 2 step
Minimal supported screen width with 100% scaling is 1024px.
Anything lower than 1024px is going into 2 horizontal columns and not much usable.
Right now it is not possible to create a usable interface with lower resolution, if you have any ideas - tell me. 

https://github.com/user-attachments/assets/cefb646b-24cd-41d5-a63a-19c28245305b

### DIFF
--- a/bases/rsptx/assignment_server_api/assignment_builder/src/components/routes/AssignmentBuilder/components/exercises/components/CreateExercise/components/MultiChoiceExercise/MultiChoiceExercise.tsx
+++ b/bases/rsptx/assignment_server_api/assignment_builder/src/components/routes/AssignmentBuilder/components/exercises/components/CreateExercise/components/MultiChoiceExercise/MultiChoiceExercise.tsx
@@ -131,6 +131,7 @@ export const MultiChoiceExercise: FC<BaseExerciseProps> = ({
           <MultiChoiceOptionsWrapper
             options={(formData.optionList || []) as OptionWithId[]}
             onChange={handleOptionsChange}
+            statement={formData.statement || ""}
           />
         );
       case 2:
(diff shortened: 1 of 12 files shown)


bnmnetp:
@ascholerChemeketa - yes the statement is authored in step 1 and you can safely hop back and forth.  The issue that was raised was that hopping back and forth is too much cognitive load on the question writer.  so having it displayed  on step 2 would reduce that load.

I haven't written enough questions where I felt like there was a need to remind myself what the question was when I was coming up with options, so I'm not a good one to judge.  But my guess is that the next bit of feedback for this design will be that there isn't enough room to write the options.

In my head, I would always display the statement above the options and let the panel for the options grow as large vertically as it needed to.  If one had to scroll up to move on to step 3 that seems acceptable.

ascholerChemeketa:
I have authored questions where the statement has a program with non-trivial amounts of code that takes up a significant amount of space. So I can see running into these spacing issue on those. But for many questions, the statement is going to be nice and short.

I definitely don't want to have to scroll past the preview to get to what I am trying to author in step 2. That feels like punishing me if I do remember what I just wrote in step 1.

Maybe a show/hide toggle on the preview panel in step 2? Especially if the state of that is persistent, each author can leave it set to whatever they feel works best.

barbarer:
I prefer entering the statement and options on one page as Brad says due to cognitive load if they are split.  If that is the case I would prefer the statement above the options. If we are entering the statement on one page and the options on another then I would put the statement as a reminder under the options on the second page. That way if the statement is long it doesn't take focus away from the options.

barbarer:
I like Beryl's idea to make it collapsible.  In that case I would put the statement first.

PR #932, opened 2025-07-31
Trim each line's right whitespace before comparing, for iotest outputs

skiadas:
Currently iotests are compared by first trimming the line end of the overall output block, even if it is a multiline output. I think the more desired behavior would be to trim each individual line of that multiline output, (and to only right-trim, though I am open to either).

### DIFF
--- a/bases/rsptx/interactives/runestone/activecode/js/livecode.js
+++ b/bases/rsptx/interactives/runestone/activecode/js/livecode.js
@@ -571,9 +571,10 @@ export default class LiveCode extends ActiveCode {
         let passedTests = 0;
         // we will stop on the first error
         this.errinfo = null;
+        const trimLines = (s) => s.split("\n").map( s => s.trimEnd()).join("\n")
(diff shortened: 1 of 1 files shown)


bnmnetp:
@ascholerChemeketa ??

bnmnetp:
I'm not sure where this is going, but given that it is in `livecode.js` this is not going to be helpful for the in-browser python.

If you want to expand iotest support for more languages it should be refactored to live in the activecode base class.

skiadas:
Hm right this just fixes the current behavior for iotests. I'll look into understand the relation between livecode and activecode, I just looked at where the iotest work was being done. Since this is meant to also work with languages that are not supported in-browser, can it still all be done in activecode?

ascholerChemeketa:
IOTests currently only exist for jobs based languages. Browser based would require making sure stdin/out can be hijacked and that the program can be executed multiple times.

Getting that all working would be nice, but feels like a separate project.

PR #937, opened 2025-08-05
Fix issue 871 - Feature Request Implement editor for selectquestion

morozov-av:
Hi, @bnmnetp!
The loading issue with the selectquestion preview seems to be happening because the question tries to asynchronously load other questions, which isn't currently supported in the assignment builder.

I also had a few questions:
1. Where should the data-points value come from?
2. What format is expected for data-togglelabels?


bnmnetp:
I think its fine for now if the whole assignment preview shows things correctly.

PR #943, opened 2025-08-12
Fix ssue 916 - New Assignment Builder: default value for number of activities in Readings not saved to show in assignment

### LINKED ISSUE #916: New Assignment Builder: default value for number of activities in Readings not saved to show in assignment
In the new assignment builder, in the choose readings page, if you do not change the default value under the "# required" column, it shows it as "None activities required" in the student view of the assignment (so I think some variable is not being set): 

Image

If you change the value under that column, then it does show the correct value in the student view of the assignment.

morozov-av:
Fix #916


bnmnetp:
fixes #916

PR #956, opened 2025-08-20
issue-953 Add "released" toggle to main assignment builder page

### LINKED ISSUE #953: Add released toggle on assignment builder main page
The assignment table has a boolean column for "released"

Please add a switch (next to the visible) switch that updates the released field.  The default for a new assignment should be released is true.

morozov-av:
Fix #953


bnmnetp:
Thanks for the quick turnaround on this !!

Issue #945, opened 2025-08-12
New assignments not being automatically graded in gradebook


bhoffman0:
I created assignments in the base course csawesome2 using the new assignment builder, but teachers are reporting that the student progress is not being automatically updated in the gradebook for these new assignments. 
I tested this with a test student account BerylTest1 in the course csawesome-demo. I did 20% of the 1.2 Variables assignment (for example csawesome2__chooseType1) but it shows as Not Started in the old and new grader pages. For assignment 1.1, it showed Not Started until I ran "Score questions and update totals" in the Grader and then the gradebook showed the correct grade. In old assignments, we did not need to click on the grading button, it would automatically update the gradebook.

<img width="950" height="444" alt="Image" src="https://github.com/user-attachments/assets/a33fdcdc-b752-402e-b496-a6c54717b3b4" />

<img width="944" height="320" alt="Image" src="https://github.com/user-attachments/assets/042665d7-cf83-41bf-a157-e76d1623bd66" />

<img width="908" height="468" alt="Image" src="https://github.com/user-attachments/assets/32e8e6d6-d723-4e74-9366-20d8a5ce2751" />

bnmnetp:
I think this might be related to a change Andrew made, if you redo the experiment but release the grades before your test student does any work, then I think the update to the gradebook will happen.

I'm wondering if it might be time to get rid of the release grades button all together.

bhoffman0:
Another teacher just reported this to me as well., where they have to click on the Score Questions button in the old grader before students and they can see the grades. 
Can we go back to having the gradebook update automatically instead of having to click on the Score Questions and Update Totals button? 
@ascholer is this due to a change that you made?

bnmnetp:
@bhoffman0 -- We need more data on why this is happening.  I have tried several times and I am getting the expected behavior of the question getting scored and the gradebook getting updated.

bhoffman0:
Ok, I will ask some more teachers and do more experiments too. I forwarded the email of the teacher having problems to you.
Thanks for looking at this.

bnmnetp:
Thanks for your help in tracking down where the issue lies.

bnmnetp:
@bhoffman0 -- Another wrinkle that I just thought of.  If an instructor creates a question from the assignment builder and adds it to the Chapter and subchapter, then that may be counted as part of the number of questions for a reading assignment.  I don't think that is right.

[...]

PR #986, opened 2025-08-29
issue-945 Refactor ID sanitization across multiple components

morozov-av:
1. Created Unified Sanitization Utility
2. Updated All Preview Components
3. Updated BaseExerciseSettingsContent.tsx to validate exercise names in real-time using the validateIdName() function. Now all exercise creation forms show specific validation errors.

**Allowed characters**:
Letters (a-z, A-Z)
Numbers (0-9)
Hyphens (-)
Underscores (_)

**Not allowed:**
Spaces (completely forbidden)
Special characters (@, #, $, %, etc.)
Unicode characters
Starting with numbers, hyphens, or underscores

issue #945

### DIFF
--- a/bases/rsptx/assignment_server_api/assignment_builder/src/utils/sanitize.ts
+++ b/bases/rsptx/assignment_server_api/assignment_builder/src/utils/sanitize.ts
@@ -2,7 +2,7 @@ export const sanitizeId = (name: string, fallback?: string): string => {
   const nameToUse = name || fallback || `generated_${Date.now()}`;
 
   let sanitized = nameToUse
-    .replace(/[^a-zA-Z0-9\-_]/g, "_") // Replace invalid characters with underscore
+    .replace(/[^a-zA-Z0-9\-_]/g, "_") // Replace invalid characters (including spaces) with underscore
     .replace(/_+/g, "_") // Collapse multiple underscores to single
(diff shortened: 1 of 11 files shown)


bases/rsptx/assignment_server_api/assignment_builder/src/utils/sanitize.ts
```diff
@@ -0,0 +1,36 @@
+export const sanitizeId = (name: string, fallback?: string): string => {
+  const nameToUse = name || fallback || `generated_${Date.now()}`;
+
+  let sanitized = nameToUse
+    .replace(/[^a-zA-Z0-9\-_]/g, "_") // Replace invalid characters (including spaces) with underscore
+    .replace(/_+/g, "_") // Collapse multiple underscores to single
+    .replace(/^_+|_+$/g, ""); // Remove leading and trailing underscores
```
bnmnetp, line 7: Many questions from a pretext book will have a name that contain multiple underscores.  For example: `csawesome2__foo_bar-untit-3`

If the author included edition numbers then the name would be `csawesome2_2_foo_bar-unit-3` where the 2 indicates a second edition.

morozov-av:
> Thanks @morozov-av for making it all consistent and enforcing it the UI as well.
> 
> I think we should allow double underscores.

Then it's even much simpler. I'll just clean out all forbidden characters and won't control the number of underscores in a row

PR #1020, opened 2025-10-02
Livecode: better handling of trailing newlines and pre formatted output boxes

ascholerChemeketa:
The updated whitespace trimming results in student answers that are missing a trailing newline always being wrong.

PreTeXt always generates a newline at the end of the `<output>` block of an `<iotest>`. I don't necessarily love that, but it is consistent with other PTX behavior when processing code like blocks. So in the screenshot below, the expected answer is:
```
4 5

```
With a trailing newline.

The provided answer is:
```
4 5
```
Without a newline. 

The rendered output gives no indication that there is an issue. Even wrapping the outputs in `<pre>` tags won't make the issue visible. 

image

@skiadas This will mean students never get answers wrong for missing or extra trailing newlines. But I think that is much better than the current behavior. Do you see a reason to ever mark those wrong? If so, how do you think we should make the issue more visible to students?

### DIFF
--- a/bases/rsptx/interactives/runestone/activecode/js/livecode.js
+++ b/bases/rsptx/interactives/runestone/activecode/js/livecode.js
@@ -571,22 +571,24 @@ export default class LiveCode extends ActiveCode {
         let passedTests = 0;
         // we will stop on the first error
         this.errinfo = null;
-        const trimLines = (s) => s.split("\n").map( s => s.trimEnd()).join("\n")
+        // trim any trailing whitespace so invisible extra newline doesn't fail test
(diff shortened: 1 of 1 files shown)


skiadas:
I don't think there's any particularly good way to account for trailing whitespaces given all the various states that the corresponding expected/actual blocks go through, so I am perfectly ok with not counting anything like that as an error. If it breaks any of our tests then we'll just adjust them.

I see the screenshot is mostly single line results, are there any changes to multi-line results? Are internal end-of-line characters treated the same? Looking at the diff sounds like they work the same as before.

PR #1021, opened 2025-10-03
(Synchronous) PI Text Chat Redesign

ibrahimmoazzam:
This PR contains the following design changes to the synchronous PI text chat:
- New UI (HTML + CSS) for the text chat discussion panel, closely following Figma designs approved by Barb
- Refactored some JS code to reduce redundancy
- Fixed minor capitalization errors

The new CSS has been added to the same stylesheet that I previously created for the instructor's PI interface. Additionally, I can emulate this new design for the async chat if Barb is satisfied with it.


bnmnetp:
I don't see any reason why the poetry.lock file for rsmanage should be part of this PR.  Especially since it has a merge conflict.

PR #1050, opened 2025-11-01
Parsons help combine blocks bugfix

ascholerChemeketa:
Existing test for language is invalid, sometimes causes error midway through combining blocks which breaks the problem.

To reproduce, start with a Parsons that has a defined language. Move a block that should be heavily indented to the top of the solution area (to get past test on line 88). Fail three times, then hit help.

@sean-fitzpatrick - ~this may be part of your problem... but this bug only affects blocks in Parsons with language != "natural" or "math". Which makes me think you have something else going on~ Duh, the test itself is what breaks. Pretty sure this is your problem.

### DIFF
--- a/bases/rsptx/interactives/runestone/parsons/js/parsonsBlock.js
+++ b/bases/rsptx/interactives/runestone/parsons/js/parsonsBlock.js
@@ -92,8 +92,8 @@ export default class ParsonsBlock {
                     );
                     // todo: if language is natural or math then don't do this
                     if (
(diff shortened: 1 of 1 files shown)


sean-fitzpatrick:
In my case there is no language set in the source; the blocks contain text and some math. 

No indentation, and we're seeing trouble without use is the help button. 

I did have the adaptive attribute set to yes, and could try removing that. But it sounds like there is some more fundamental issue?

PR #1062, opened 2025-11-05
Html manifest

bnmnetp:
For testing

- [ ] Solve issue with the way some characters are encoded on conversion


bnmnetp:
Abandoning this in favor of a cleaner approach where we build an xml representation of the question that can be easily created from JSON and easily converted to JSON.

PR #1065, opened 2025-11-07
Assignment builder: allow search to match exercise qnumber field

ascholerChemeketa:
Since the `qnumber` field is the user visible title of an exercise in PTX world, it would be useful to be able to search against those in the assignment builder assignment list.


bnmnetp:
One line fix.  will be out in tomorrow's release.

PR #1069, opened 2025-11-11
Multigrader manual grading updates grading method

ascholerChemeketa:
This makes the multigrader update "autograded" status to "instructor graded" like the single view.


bnmnetp:
Thanks for the fix.

PR #1071, opened 2025-11-13
Send reading score updates when all activities are complete on page load

ascholerChemeketa:
Fixes issues where doing activities on assignment page before viewing reading page would not trigger reading score update.

### DIFF
--- a/bases/rsptx/interactives/runestone/common/js/bookfuncs.js
+++ b/bases/rsptx/interactives/runestone/common/js/bookfuncs.js
@@ -157,20 +157,26 @@ class PageProgressBar {
             subchapterprogress.max = 100;
             subchapterprogress.value = value;
         }
-        // Handle the case where there are no interactive objects on the page.
-        // but we still need to give the student points for reading the page.
-        if (this?.assignment_spec?.activities_required === 0) {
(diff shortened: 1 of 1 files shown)


bases/rsptx/interactives/runestone/common/js/bookfuncs.js
```diff
@@ -157,11 +157,13 @@ class PageProgressBar {
...
             subchapterprogress.value = value;
         }
-        // Handle the case where there are no interactive objects on the page.
-        // but we still need to give the student points for reading the page.
-        if (this?.assignment_spec?.activities_required === 0) {
+        // If the user has completed all activities, send the reading score.
+        // This handles the case where there are no activities on the page or
+        //  where the user completed activities on the assignment page and now
+        //  is viewing the reading page.
+        if (this?.possible === this?.total) {
```
bnmnetp: What if the number of activities required is less than the total possible?  The 80% rule we typically create.  I think we might want to use the number required from the assignment_spec here???

ascholerChemeketa:
Yes, that does make sense. I always require everything, didn't even realize that requiring less is a thing.

Force pushed with change to compare completed activities to required.

PR #1074, opened 2025-11-20
Add Kotlin capability

dmusican:
Here are updates needed to get Kotlin working, following on the updates that were done to the [Runestone Jobe repo](https://github.com/RunestoneInteractive/jobe/pull/2) and to [PreTeXt](https://github.com/PreTeXtBook/pretext/pull/2719).  Some particular notes:

- While CodeMirror does have a mode for [Kotlin](https://codemirror.net/5/mode/clike/index.html), it does not appear support a MIME type for Kotlin. So in `activecode.js` I used the Java MIME type for Kotlin code. Should be close enough.

- To install Kotlin into the container that runs Jobe, I didn't do the easy/obvious approach of installing from Ubuntu's apt repos because the version there was very out-of-date. Instead, I added code to the Dockerfile to download and install directly from JetBrains.

- There was a major conflict between `pretext` and the package `dash-bootstrap-components`, which was listed in the developer set of packages. So I removed `dash-bootstrap-components`, and committed both the new `pyproject.toml` as well as an updated `poetry.lock` file.

I'm happy to answer questions or make changes, so let me know how I can help. Thanks!


bnmnetp:
Thanks Dave!

I just pushed 7.10.0 to the CDN

PR #1075, opened 2025-11-22
Better FastAPI server error pages

ascholerChemeketa:
This sends HTML error pages from the FastAPI servers to clients that accept `text/html` instead of JSON. 

Original goal was to improve the message LTI1p3 links were making when Redis connection failed. But the natural point to return the right page also gets used by author/book/assignment. So, as implemented it will affect all of the FastAPI servers. Which is probably a good thing assuming you agree with the approach.

Error page includes message and timestamp corresponding to ticket:

image


bnmnetp:
Looks good, works as advertised!  Thanks.

PR #1077, opened 2025-11-25
Documentation for new programming languages

dmusican:
Here's a documentation page for the developer docs that describes the process for adding a new programming language.  I describe how to make the jobe change as a standalone piece, how to make the PreTeXt update (again as an isolated update), and then how to do the Runestone update once those two are done.

The process I went through when I added Kotlin was a bit more involved in this, because I didn't know if any of it was going to work and I was exploring. So I set things up locally so I could test the entire update from bottom to top. That's more challenging to do. What I document here is the process that I ultimately did for submitting the PRs and testing each piece in isolation, but I left out the trickier bits regarding doing local integration testing.

I'm happy to update, make clarifications, etc; let me know.


docs/source/new_language.rst
```diff
@@ -0,0 +1,122 @@
...
+
+Once you have your PreTeXt development environment working, test it before making any changes. In particular, start with a Runestone book that has active code within it, and change it to your new language. An easy example is the PreTeXt sample book; look in the file ``rune.xml``, which has a variety of examples. Within the ``<program>`` tag, change the ``language`` property to have your new language. Of course, this won't work yet, but this will help you see what you're looking to test. Once you have made that change, build the book at the command line::
```
bnmnetp: The preferred example is the runestone sample book as I use that for testing frequently

dmusican:
Update made --- I think this is what you had in mind. Let me know if not and I'll take another whack at it.

bnmnetp (approved):

bnmnetp:
Thanks!!

PR #1082, opened 2025-12-10
Shortanswer mathjax updates

ascholerChemeketa:
* Fix bug that always assumes mathjax is in use
* Adds label and aria information to rendered output
* Only shows render box once there is something to render

I'll do a separate PR for PreTeXt so `data-mathjax=""` isn't always added to every shortanswer authored there. But this is needed for that to have any effect.

### DIFF
--- a/bases/rsptx/interactives/runestone/shortanswer/js/shortanswer.js
+++ b/bases/rsptx/interactives/runestone/shortanswer/js/shortanswer.js
@@ -26,17 +26,17 @@ export default class ShortAnswer extends RunestoneBase {
             this.question = this.origElem.innerHTML;
             this.optional = false;
             this.attachURL = opts.attachURL;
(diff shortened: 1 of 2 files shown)


ascholerChemeketa:
Do you want full JQuery swapouts? 

I try to swap it out in code I touch. But I'm leary of full file replacements incase Copilot biffs something and I don't notice.

I say put it in agentic mode and turn it loose. Then, if there are bugs (or it [deletes your D; drive](https://www.tomshardware.com/tech-industry/artificial-intelligence/googles-agentic-ai-wipes-users-entire-hard-drive-without-permission-after-misinterpreting-instructions-to-clear-a-cache-i-am-deeply-deeply-sorry-this-is-a-critical-failure-on-my-part)) we can blame it.

ascholerChemeketa:
Added a commit to remove JQuery

PR #1084, opened 2025-12-12
Interactives: remove jquery and update spacing

ascholerChemeketa:
Removes jquery from many interactives. It leaves activecode and parsons alone. Those two are complex enough they deserve special attention for testing all the edge cases when jquery is removed. (Spent too long trying to figure out what bug got introduced into polls when swapped out JQuery on them. Eventually realized the bug was already there in the JQuery version, but manifested slightly differently in the new version. Worried about things like that with ac/parsons.)

Normalizes the spacing and better handles the spacing of PTX that is injected into the RS elements.

This needs to be in place to see full effect:
https://github.com/PreTeXtBook/pretext/pull/2743

### DIFF
--- a/bases/rsptx/interactives/runestone/parsons/css/parsons.css
+++ b/bases/rsptx/interactives/runestone/parsons/css/parsons.css
@@ -352,7 +352,6 @@
     border-left: 0;
     border-right: 0;
     border-radius: 0;
-    margin: 10px 0;
 }
 .parsons .parsons-text {
     max-width: var(--contentWidthRegular);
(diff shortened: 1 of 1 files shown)


ascholerChemeketa:
BTW: I tested all the JQuery conversions on the sample book and one of my books using a local RS server. Everything seems to work, but that obviously isn't exhaustive testing. The parts I would be most worried about are ajax calls, especially binding of callbacks. That was where the existing poll bug was (went back in time to fix it and then redid the JQuery conversion).

bnmnetp:
Seems to work fine for me too.  I'm going to merge.  This is a good time to merge this as traffic has dropped off significantly for the holidays.

PR #1100, opened 2025-12-30
feature-872 Add Data Files management to ActiveCode exercises

morozov-av:


bnmnetp:
This is looking good!

I think the one thing we are going to want to add is the ability to allow the student to see a bit of the data file.

Maybe we can do that as part of the instructions.  Add a check box for the datafiles to let the student see the first few lines...

bnmnetp:
I think I will merge this as is so we can get more feedback from others.

PR #1101, opened 2026-01-01
Two way navigation

DavidCooperWCU:
I added text links for assignment navigation to  
1. be able to go back a page if there is a page behind
2. have text saying that you are at page 1 of the assignment when you are.
3. have a link to return to page 1 of the assignment if you have navigated away from the assignment while reading it.

These changes are all in bookfuncs.js


bnmnetp:
Thanks, this should go live over the weekend.

PR #1102, opened 2026-01-01
Course creation: fix double handler for invoice me checkbox

ascholerChemeketa:
There are two copies of the event handler for `invoice-me`. The one on line 248 appears to have the correct logic. This one is reversed.

Currently if you click that button, the confirmation triggers twice, and the second one turns off the radio when you hit "OK". So it is impossible to offer RS money!


bnmnetp:
Good catch, thanks.

PR #1104, opened 2026-01-02
Hparson math rendering fixes

ascholerChemeketa:
Fixes new seeming issue with language detection since JQuery removal.

Missing `data-language` produces `null` now, not `undefined`. Something to watch for in future conversions. I intentionally changed the contruct to `this.language == null` using `==` since `null` does == `undefined` but they are not `===`.

Second commit restricts the mathjax scope and adds a bit of a delay. Especially with multiple problems to render (but sometimes even without them), I can produce unrendered math even before the commit that removed JQuery and introduced the new issue. The fix is kind of ugly, but I don't know how to more gracefully tap into the work that the imported microparsons is doing.

### DIFF
--- a/bases/rsptx/interactives/runestone/hparsons/js/hparsons.js
+++ b/bases/rsptx/interactives/runestone/hparsons/js/hparsons.js
@@ -195,9 +195,12 @@ export default class HParsons extends RunestoneBase {
             });
 
             if (window.MathJax && MathJax.typesetPromise) {
-                MathJax.typesetPromise().then(() => this.simulateSolution());
+                MathJax.typesetPromise([blocks]).then(() => this.simulateSolution());
             }
-        }, 0);
+        }, 100);
+        // Ugly... but there are timing issues with block rendering in microparsons and MathJax
+        // most noticeable if multiple hparsons problems are on the same page
+        // this delay seems to help avoid issues
     }
 
     /*


bases/rsptx/interactives/runestone/hparsons/js/hparsons.js
```diff
@@ -195,9 +195,12 @@ export default class HParsons extends RunestoneBase {
             });
 
             if (window.MathJax && MathJax.typesetPromise) {
-                MathJax.typesetPromise().then(() => this.simulateSolution());
+                MathJax.typesetPromise([blocks]).then(() => this.simulateSolution());
```
bnmnetp: I wonder if we shouldn't somehow be using `this.queueMathJax(myElement)` as we do in other components to ensure that we are serializing all of the various requests to mathjaxify the math.

Not 100% sure how we would know they are all done other than to check that the queue is empty and has been empty for some small period of time.

ascholerChemeketa:
Let me try that real quick...

ascholerChemeketa:
Leaving this for @geoff-cox to refer to, but otherwise abandoning work on it.

geoff-cox:
@ascholerChemeketa @bnmnetp 

I started thinking about restoring math mode in hparsons, but before getting too far, I wanted to discuss some details.

As I understand it, micro-parsons is now effectively abandoned upstream, and RS has forked it solely to support hparsons. At this point, micro-parsons does not appear to have any independent consumers within RS beyond hparsons itself.

The immediate technical issue I am addressing is that hparsons currently relies on rendered block text (including MathJax output) to identify and grade blocks. This is fragile and is the root cause of recent math-mode failures. The correct fix requires treating block identity as a stable ID (source index) and passing that information cleanly from micro-parsons back to hparsons. I am considering refactoring both sides to make “block indices” the canonical representation for answers, storage, logging, and grading.

Given this,
1. Is micro-parsons only used by hparson?
2. Should micro-parsons continue to exist as a separate forked project, or does it make more sense to treat it as an internal component of hparsons?
3. Would it be reasonable to co-locate it into RS to simplify integration and maintenance?

My short-term plan is to make the minimum changes necessary to stabilize math mode in hparsons, but I want to ensure those changes align with how you plan to use this code going forward. Any guidance would be welcome.

bnmnetp:
I **think** this can be closed as I think #1117 incorporates some of the changes, and I added queueing of the blocks to PR before I merged.

sean-fitzpatrick:
That sounds like I should rebuilt, then.

PR #1141, opened 2026-02-06
Fix CodeTailor Java version: : use JOBE to compile Java code in book_server

xinyinghou:
- CodeTailor runs Java tests automatically in book_server_api, but Java test runs fail on the platform with: Error while running Java tests: [Errno 2] No such file or directory: 'javac'.
- This update uses the JOBE Java compiler.
- Also fixes formatting issues in existing CodeTailor files.


xinyinghou:
Thanks, I have switched to JOBE and now using the book_server-associated `rsproxy.py`. I tested and it works on my server.

PR #1158, opened 2026-02-23
Datafiles: fix where filename is accessed from in constructor

ascholerChemeketa:
This bug appears to date back quite a ways.

The `data-filename` is not copied from `orig` to the replacement container that is created by Runestone. Thus the filename is never correctly set.

This impacts livecode's that try to access that filename to use in sending the file to Jobe, but only when the datafile is on the same page. When the datafile is coming from the database, things are fine.


bnmnetp:
This looks like you are setting it in `orig.dataset` the code as is sets the filename replacement container. 

But, I'll admit that with just a quick glance at the code I don't even see where `this.fileName` is getting set.  Looking at the history there was a line: `this.fileName = orig.dataset.filename || null;` but it appears to have ben removed in the PR where you added the `parseBooleanAttribute` method.

Issue #1145, opened 2026-02-10
Feature Request  Remember the last column chosen to sort the assignment list


bnmnetp:
Instructors have requested that the list of assignments remember the column used to sort the assignments.

Using localstorage to remember this should be totally acceptable.

bnmnetp:
Implemented in commit 70ff1e55 — the assignment list now persists the sort column and direction in localStorage (`assignmentList_sortField` / `assignmentList_sortOrder`) and restores them on load, with unit tests covering the persistence. Closing.

Issue #814, opened 2025-06-23
On assignment page view - add creation date and allow sort on that


barbarer:
This was a request from the PI workshop in June 2025.

In the new assignment builder on the page that shows all the assignments add the creation date and allow the user to sort in reverse order so that the most recently created are on top.

bnmnetp:
There is already a due date column that is sortable.  I think adding another date would just make it more confusing.

bnmnetp:
Closing — the assignment list already has a sortable due-date column, and adding a second date column risks more confusion than it resolves.

If sorting by creation date specifically is what you need (e.g. to find the assignments you just made), open a focused issue and we can look at it as a sort option rather than as a new column.

Issue #466, opened 2024-01-26
On assignments allow the instructor to set a day/time for it to be make visible


barbarer:
Right now I have to go into the instructor page and make my assignments for lecture visible just before or during lecture.  It would be nice to have them automatically become visible at set day/time.

jjjonesj:
I can take on this issue!

bnmnetp:
I'm not sure what the best way to handle this is...

Right now we have a simple switch
If we add a date/time then we can just check that before showing the assignment.  But what if you want to hide the assignment again, like after an exam?  Just set the  visible date to something far in the future?

It would be nice to not have both the switch and the date, and I really don't want to have both a show date and a hide date.

@barbarer @bhoffman0 What is the right flow here that will not confuse people?

barbarer:
I agree that we need a show date and time and optionally a hide again date
and time.

Dr. Barbara Ericson
Associate Professor, School of Information
University of Michigan

jjjonesj:
Hypothetically, wouldn't the hide date fall under the due date system? Or would we still have the same issue of having the calendar and switch in that case as well? If we want things to be automatic, the due date system would close the assignment once the specific time passes while the visibility just controls when it is seen, maybe?

bnmnetp:
Every instructor has their own tolerance for how late to accept work.  So you can't just hide the assignment on the due date or even X days after the due date.

In my case I said I would accept late work until I got around to grading it.  So sometimes it was days and sometimes a week or more.

bnmnetp:
My feeling is that we eliminate the switch, replacing it on the main panel with a date/time picker for when to make the assignment visible.  It should default to the current time.

Then inside the "More Options" panel we add a "Hide After " date / time picker.

The logic in the chooseAssignment and doAssignment endpoints will need to be updated to use the dates if they are there, but fall back on the `is_visible` switch for old assignments created before we add this.

We don't have a background task to monitor the  time to make something visible or invisible we just control what can be seen in the two endpoints we have for choosing and doing an assignment.

jjjonesj:
I see now, that makes sense. I'll have to explore the files more for the "More Options" and to better understand eliminating the switch over the weekend.  In the meantime, if we get rid of the switch for the new date system, how do we keep it partially intact for old assignments?

bnmnetp:
The visibility of an assignment is represented in the model/database by the `is_visible` column.  We will shift this to use `visible_on` or some other better name to indicate when an assignment should be visible.
[...]

PR #1160, opened 2026-02-24
Issue-814 Issue-466 Issue-1145

morozov-av:
#1145   isue Implement sorting functionality in AssignmentList with localStorage persistence
#814  On assignment page view - add creation date and allow sort on that
#466  On assignments allow the instructor to set a day/time for it to be make visible

### DIFF
--- a/bases/rsptx/assignment_server_api/assignment_builder/src/components/routes/AssignmentBuilder/components/edit/VisibilityControl.tsx
+++ b/bases/rsptx/assignment_server_api/assignment_builder/src/components/routes/AssignmentBuilder/components/edit/VisibilityControl.tsx
@@ -2,7 +2,7 @@ import { RadioButton } from "primereact/radiobutton";
 import { Control, Controller, UseFormSetValue } from "react-hook-form";
 
 import { Assignment } from "@/types/assignment";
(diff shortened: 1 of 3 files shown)


bnmnetp:
@pearcej So, the visible from ... to setting is meant to address the use cases you are talking about:

1. PI assignments
2. exams

We have to balance the fact that the vast majority of users of Runestone do NOT use LTI.  So although LTI users find pain in having to manage some things in both places, most users only manage things on the Runestone side.  I'm not sure if there is a good solution to this.

ascholerChemeketa:
Jan said in this thread:

> I think it would be useful to be able to launch an assignment in Runestone that is not visible. I thought we used to be able to do this, but it would be especially useful for quizzes or tests or even peer instruction questions.

Presumably anyone doing that would want grades recorded.

bnmnetp:
@bhoffman0 @barbarer 

What do you think?  @barbarer I know you have wanted this for some time.

Particularly take a look at the image from @morozov-av  -- Note that this provides backward compatibility and keeps the functionality of (and backward compatibility) marking something visible / not visible going forward.

If things look good this week with further testing I hope to merge it and have it available next weekend.

PR #1162, opened 2026-02-25
Fix async peer LLM token lookup, duplicate token bug, and improve async message view

sethbern:
This PR fixes a few issues with the async peer LLM feature and cleans up the instructor view.

There was a duplicate token issue on the add token screen, the custom provider name input happened to share the `token-input` CSS class with the actual key field, so the submit handler was collecting it as a second token every time. The renamed class fixes this.

Fixed it so all messages from a student show up in order instead of just their last one in async peer when there is no llm. Also the LLM prompt has been updated.

### DIFF
--- a/bases/rsptx/interactives/runestone/common/css/runestone-custom-sphinx-bootstrap.css
+++ b/bases/rsptx/interactives/runestone/common/css/runestone-custom-sphinx-bootstrap.css
@@ -393,27 +393,6 @@ div.container {
   background-color: #f5f5f5;
 }
 
-/*Dark mode-force dropdown menus to black background with white text */
-[data-theme="dark"] .dropdown-menu {
-    background-color: #000000;
-    color: #ffffff;
-}
-
-[data-theme="dark"] .dropdown-menu > li > a,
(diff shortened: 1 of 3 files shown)


sethbern:
When I force `window.PI_LLM_MODE = true` in the console on the deployed server and trigger a chat, the `get_async_llm_reflection` endpoint returns {"ok": false, "error": "missing api key"}. On my local setup the same call returns the key correctly and the LLM responds fine. So the issue seems to be that the deployed server isn't finding the API key.

sethbern:
@bnmnetp I just tried on a separate device with a fresh Runestone build and the feature worked correctly there, so it seems like a version mismatch somewhere on the deployed server might be the culprit. Could something be out of sync between what's deployed and the latest merged code?

bnmnetp:
I see one api token for that course for the openai provider.  It looks like it has never been used as the last_used column is null.

sethbern:
Got it, thanks for tracking that down. Whenever you get a chance, would it be possible to add FERNET_SECRET to the runestone service? That would let me test the current version while I work on migrating the async PI implementation over to FastAPI. And I really appreciate your help and support on this!

Issue #612, opened 2024-12-20
Textbook - When the Dark Mode is on, the text in the navbar dropdown menus becomes invisible.


bnoggle:
Description: When the Dark Mode is on, the text in the navbar dropdown menus becomes invisible. The menu options are shown when the user hovers the cursor over an option.

Steps:

1. Log in as a user.
2. The Course Home page will appear. Click the textbook link (“Runestone Interactive Overview”).
3. Click the gray user icon located at the upper right corner of the textbook page and open the user dropdown menu.
4. Activate the Dark Mode by turning on the toggle switch at the bottom of the dropdown. 

Expected Results:

The Dark Mode activates and all elements and text are displayed.

Actual Results:

The Dark Mode activates, and the menu option text in the dropdown menus (Search, User, and Help dropdowns) becomes invisible.

Environment:

Device - HP Pavilion (laptop)

Screen size - 15.5”

Display resolution - 1366 x 768

OS - Windows 11 Home

Browser - Google Chrome Version 131.0.6778.109 (Official Build) (64-bit)

![Dark Mode](https://github.com/user-attachments/assets/5b9ae2a0-7ab7-4e76-9c23-aa6e4247f4bf)

PR #1164, opened 2026-02-26
Issue-612 Fix dark mode dropdown menu and chapter navigation text visibility

sethbern:
Fixes issue #612

The textbook page has a user-controlled dark mode toggle in the navbar that sets data-theme="dark" on the <html> element when activated. There's no forced dark mode, the CSS just wasn't handling that attribute on the affected elements.

This adds [data-theme="dark"] rules for:
- Dropdown menus (Search, User, Help) 
- The chapter/jump-to navigation select in user-highlights.css

### DIFF
--- a/bases/rsptx/interactives/ptxrs-bootstrap.less
+++ b/bases/rsptx/interactives/ptxrs-bootstrap.less
@@ -12,6 +12,8 @@
     --bodyFont: var(--body-text-color, #000000);
     --tooltip: #ffffff;
     --grayToWhite: #333333;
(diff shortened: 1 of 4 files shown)


bnmnetp:
I pulled this down and tried rebuilding the overview book with it.  But it does not change the styling of the menus.

bnmnetp:
Books will need to be rebuilt in order to get the new css and javascript.  I did not do a mass update of the books over the weekend as this was really the only change to the javascript and didn't seem urgent as it has been this way for quite a while.

PR #1171, opened 2026-03-06
Created Initialization Script + Documentation

conzty01:
I created an initialization script that can be used to get the project set up and running locally with an interactive wizard instead of the manual steps outlined in the [`running.rst`](https://docs.runestone.academy/en/latest/running.html) documentation. This script has been tested in a Windows WSL environment but has been built to function in UNIX environments. Unfortunately, I don't have the resources to test in Linux or MacOS at the moment, but can work on getting that done if necessary.

This script essentially runs through the steps outlined in the `running.rst` documentation in an interactive way-- prompting the user for paths or values. It detects when the script has been run more than once and backs up any .env files in the root directory and provides options for how to handle an existing database:

1. Skip initialization and continue
2. Reset database
3. Exit script and manually intervene

I'd appreciate some feedback on the portions regarding adding a course as I ran into difficulties even with the manual steps in that area. I'm not sure if the docs are out of date, or I am just not understanding something but I'd be happy to take another look and alter the prompts so that this part of the process makes sense for a new user.

This PR was created with the assistance of GitHub Copilot.


bnmnetp:
Thanks @conzty01 !
Just a couple of additional fixes while testing on Linux and MacOS, and this exposed one bug in `addbookauthor` that I also fixed.  Appreciate the contribution, and your long term support of this project.

PR #1173, opened 2026-03-08
Add per-question LLM/Standard mode toggle for async peer instruction (discussion)

sethbern:
This PR gives instructors the ability to toggle async peer instruction on and off and choose which PI questions use an LLM. This is what I have so far I would love any feedback, especially with the assignment dashboard since there have been recent changes. @bnmnetp @barbarer 

- Instructors can toggle async peer mode on/off from the assignment list
    - image
- Each peer question can be set to LLM or Standard mode individually (defaults to Standard)
    - LLM option only appears when async mode is enabled
    - Edit All lets you bulk switch all questions at once
    - image
    - image
- LLM option only shows up if there's a valid API key configured
    - image

- If the API key is broken students get a fallback to standard discussion instead of an error
     - this also works if somebody is in the middle of a conversation with the llm
     - image

- Deleting an API key flips everything back to Standard and re-adding restores previous settings
- Students see a small disclaimer when chatting with the AI peer


components/rsptx/db/models.py
```diff
@@ -681,6 +681,7 @@ class AssignmentQuestion(Base, IdMixin):
     activities_required = Column(
         Integer
     )  # only reading assignments will have this populated
+    use_llm = Column(Web2PyBoolean, default=False)
```
bnmnetp: Why would this be on a per question basis?  Shouldn't this be on an assignment ?

bnmnetp:
Very good!  That is a lot of servers this PR touches, but seems to be working as it should.

Merging now so you can get on with your research.

Also, thanks for working through the options/objections on this one!

Issue #1187, opened 2026-03-24
Feature Request  Document the JSON representation for each question type.


bnmnetp:
The long term strategy for question representation should be to have all questions with a json representation.

Getting the json format documented is a good first step.

The second step would be to write code for processing the runestone manifest that could take the PreTeXt representation of a question and create the json for the database.  This will require work on both the PreTeXt side to provde the PreTeXt source as well as on our side to convert the PreTeXt source to json.

Issue #1186, opened 2026-03-24
Feature Request  remove the toggle for released from the main assignment list


bnmnetp:
We have changed the logic and this can be removed from the main list now.

PR #1191, opened 2026-03-30
issue-1186 Remove released status handling from AssignmentBuilder

morozov-av:
… AssignmentList components

#1187  Add JSON schema documentation for question_json field
#1186 Remove released status handling from AssignmentBuilder


bnmnetp (commented):
This looks good.  I like the structure.  The markdown doesn't work with our RST in the docs structure.  But I had Copilot convert it to RST as a test and it looks fine to me.  If you have your own script or tool for generating some of this and it can create RST that might be better for whenever this changes.  If not let me know and I can use the RST as I converted it.

PR #1217, opened 2026-05-16
Fix truthy string returns masking failures in unittest evaluation

xinyinghou:
Error paths in evaluate_fixed_code returned non-empty strings (e.g. "We got errors, ..."), which are truthy and caused callers checking `if not unittest_result` to treat failures as successes. Replace all such returns with False, and tighten the caller checks to `is not True` so only an actual True passes.


xinyinghou:
The `crud-tests` CI check is failing with:

ImportError: cannot import name 'consume_reset_token' from 'rsptx.db.crud' 
(/home/runner/work/rs/rs/.venv/lib/python3.10/site-packages/rsptx/db/crud/__init__.py)

This is triggered by `bases/rsptx/admin_server_api/routers/auth.py` importing `consume_reset_token` from `rsptx.db.crud`.

My changes in this PR were only ruff fixes to one file, neither touched `auth.py` or `crud`, so flagged this as it looks like a pre-existing CI/install issue rather than sth from my changes.

bnmnetp:
A rebase should fix the issues with the tests now.

PR #1228, opened 2026-06-04
Synchrnous PI fixes from feedback

sethbern:
Fixed several sync PI issues from testing feedback:
- Show real student names in the verbal group select dropdown instead of anonymous names
- Fixed premature "discuss" message after vote 1 stops
- Added line numbers to code blocks displayed in PI questions (does not touch the actual questions, just adds them visually during a PI session based on instructor feedback that it makes discussion easier)
- Improved group select placeholder text

Also, some PI pages still seem to route to web2py instead of FastAPI. I think it's because nginx serves /staticAssets/ from the web2py static directory, but if it's something else on my end let me know and I can fix it.

### DIFF
--- a/bases/rsptx/assignment_server_api/routers/peer.py
+++ b/bases/rsptx/assignment_server_api/routers/peer.py
@@ -1222,7 +1222,6 @@ async def clear_pairs(
 
 
 @router.get("/course_students")
-@instructor_role_required()
 @with_course()
 async def get_course_students(
     request: Request,
--- a/components/rsptx/templates/assignment/student/peer_question.html
+++ b/components/rsptx/templates/assignment/student/peer_question.html
@@ -108,7 +108,7 @@ Peer Instruction - {{ assignment_name }}
             <div id="group_select_panel" class="col-md-6" style="display: none;">
                 <p><strong>Who did you talk to?</strong> Click to open the dropdown, type a name to search or scroll to select.</p>
                 <select id="assignment_group" multiple class="assignment_partner_select" style="width: 95%">
-                    <option value=""></option>
+                    <option value="" disabled hidden></option>
                 </select>
                 <p style="font-size:12px; color:#6b7280; margin-top:4px;">You can select multiple people.</p>
             </div>
--- a/components/rsptx/templates/assignment/student/peer_question.html
+++ b/components/rsptx/templates/assignment/student/peer_question.html
@@ -31,16 +31,7 @@ Peer Instruction - {{ assignment_name }}
         display: none;
     }
 
-    .pi-lineno {
-        display: inline-block;
-        min-width: 2em;
-        color: #999;
-        text-align: right;
-        border-right: 1px solid #ddd;
-        padding-right: 0.5em;
-        margin-right: 0.8em;
-        user-select: none;
-    }
+
 </style>
 {% endblock %}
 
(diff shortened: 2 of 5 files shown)


bnmnetp:
When we build the nginx container we copy only the components/rsptx/templates/staticAssets folder into the nginx container.  It does not have access to the web2py files.

components/rsptx/templates/assignment/student/peer_question.html
```diff
@@ -201,5 +214,16 @@ <h2>Peer Instruction: {{ assignment_name }}</h2>
         await setupPeerGroup();
     }
     doIt();
+
+    setTimeout(function addLineNumbers() {
```
bnmnetp: Is this trying to add line numbers to code in a multiple choice question?  If the question is written properly it will already have line numbers.

bnmnetp (commented):
It does not look like this fixes any routing problems.  Can you make a list of where you think the routing has gone wrong?

components/rsptx/templates/assignment/student/peer_question.html
sethbern: Yeah this is adding in line numbers that are not rendering during PI even when the question is created with them. I believe the questions are being created properly using the assignment builder interface and it still does not show them.

sethbern:
For the routing issue, when navigating to PI from the book page, it routes to `runestone/peer/student.html` or `runestone/peer/instructor.html` instead of the correct FastAPI routes `assignment/peer/student` or `assignment/peer/instructor`. Navigating from the course home or instructor dashboard goes to FastAPI correctly. It seems like the book page links are the only thing still pointing to the old web2py routes.

bnmnetp:
Yes we need to make some updates to the master templates of the books both on the `legacy_support` branch for old RST based books. And in PreTeXt.  I'm already working on a PR to do that for PreTeXt that will encompass the PI pages as well as new login/logout/register pages.

components/rsptx/templates/assignment/student/peer_question.html
bnmnetp: Can you point me to a specific question?  It may be that we are just missing including the prism js/css files that do the coloring and line numbers.

bnmnetp: Yes, the templates for `peer_question.html` and `doAssignment.html` are missing the needed prism files for decorating code questions.   Adding those is the right fix.

sethbern:
Ok I went ahead and fixed the line numbers by adding the missing Prism CSS to static_assets.html and removing the custom JS workaround I had. It works for questions created with the new assignment builder but older questions don't show line numbers since their HTML source is missing the line-numbers class on the pre tag. Is it worth backfilling that class on existing questions or is the expectation that instructors should just update them?

bnmnetp:
When you say old questions, are they questions written in PreTeXt or are they old RST based questions?


Old RST questions required that the author used `.. code-block:: language` and then there was a parameter like `:linenos:` to make it number the lines.

sethbern:
Old RST-based questions. They use div class="highlight-python notranslate" instead of pre class="line-numbers" so Prism doesn't pick them up.

bnmnetp:
No RST used its own method for adding line numbers and coloring code.

bnmnetp:
Se my little edit to the previous comment.

sethbern:
I see then so for those RST questions without line numbers it was just an authoring choice. In that case is it best to leave those as is?

bnmnetp:
Yeah, if the authors of those questions want line numbers they can edit the book (or question) and add them.  I don't think we should impose a style over the top of something an author already has control of.

bnmnetp:
Now, time is short for RST questions.  The editor for those old questions will not be gone by the start of classes in the fall.

sethbern:
Got it, makes sense to leave it to the authors then. I also cleaned up some leftover line number CSS and JS from the dashboard template. Anything else I should do before merge?

bases/rsptx/assignment_server_api/routers/peer.py
```diff
@@ -1222,7 +1222,6 @@ async def clear_pairs(
 
 
 @router.get("/course_students")
-@instructor_role_required()
```
bnmnetp, line 1225: This should at least require a login, maybe that is implied by @with_course but it looks like a security hole to get a list of students that are in a course if it is unprotected.

bnmnetp: Nevermind...

PR #1231, opened 2026-06-09
Fix CodeTailor security vulnerabilities

aspadiyath:
- Add authentication (Depends(auth_manager)) to parsons_scaffolding endpoint
- Remove exec()-based Python test runner; replace with JOBE sandbox via evaluate_fixed_code.py
- Fetch test code from DB (question_json.suffix_code) instead of accepting it from request body
- Add clean_python_testcase() to transform Runestone unittest.gui format for JOBE compatibility


bnmnetp:
Have you tested this locally so you know it works?

aspadiyath:
Yup, I've tested locally and it works!

bnmnetp:
Great, I'll look at it a little more and get it merged.  Thanks for addressing this so quickly.

PR #1233, opened 2026-06-10
changes for gdscript browser activcode

DavidCooperWCU:
I committed the 3 changes that I needed to make gdscript work with activecode in the browser.

I didn't add the change I needed to make to package.json, and package-lock.json in order to build. If you'd like me to commit those, please let me know.


bnmnetp:
Where is the wasm file for Godot?  I think at a minimum it needs to be in the js folder with the activecode_gdscript.js

DavidCooperWCU:
The wasm file is hosted in the book. That's what we discussed over Zoom. The latest version of the wasm is in my fork of the [pretext sample book](https://github.com/DavidCooperWCU/pretext/tree/master/examples/sample-book/ext/godot-shell).

bnmnetp:
That is not what I thought we discussed.  We can't make authors of books responsible for going out and finding wasm files.  It also makes this feature unusable unless someone is aware of that.

If your code for `activecode_gdscript` correctly imports the wasm file, then it will automatically get copied into **any** book that wants to write an example like this.

DavidCooperWCU:
The JavaScriptBridge filtering is to address the largest security concern with using a Godot WASM file. 
This concern is that GDScript can call arbitrary javascript, which could potentially open up the site to known attacks.

The filtering restricts JavaScript calls that aren't specifically postMessages that have a JSON dictionary as a payload, and a known origin. As long as eval isn't used on the JSON dictionary, this should restrict calls to actual GDScript instead of allowing arbitrary javascript code to be run in the WASM.

DavidCooperWCU:
> That is not what I thought we discussed. We can't make authors of books responsible for going out and finding wasm files. It also makes this feature unusable unless someone is aware of that.
> 
> If your code for `activecode_gdscript` correctly imports the wasm file, then it will automatically get copied into **any** book that wants to write an example like this.

Okay,

The `activecode_sql` file uses:

```javascript
import initSqlJs from "sql.js/dist/sql-wasm.js";
```

I'm not sure where sql.js is hosted, but perhaps hosting it a similar way would be useful rather than having a lot of bloat in the runestone/js directory.

bnmnetp:
That is installed by `npm` which puts all of the files in the `node_modulesq folder of `bases/rsptx/interactives/`. 

When we prepare a release of the components `webpack` traces all of the imports and bundles them up into easily chunkable parts, so that we only load what we need for each web page.

What you do for gdscript should work similarly.

DavidCooperWCU:
So, I'm not sure if npm makes the most sense for this because index.html is required for this setup, but  I could look into doing something similar to the way micro-parsons-element is done and make a repository that just has the godot-shell directory in it along with a package.json file that copies the files. 

Does that make sense?

DavidCooperWCU:
Alternatively I could put a directory somewhere in `interactives/runestone` since it's really designed to work with the runestone activecode framework. I'm just not sure the best place for it. `activecode` has `js` and `css`. 

I could make a `godot-shell` folder there (in `interactives/runestone/activecode` with the wam and supporting pck, html, images, and javascript. 

Do you think that's the best place for it? If so, then I just need to figure out how to make the appropriate package.json file so that it copies the files correctly.

bnmnetp:
Definitely not the micro-parsons approach.  We are working on moving that code directly into the hparsons folder.

We need to be careful to distinguish what is the framework needed to run a godot program and what is needed by individual problems.   What is the purpose of the following:

* index.html ?
* other html files???
* pck ??
*  images?

I'm starting to get the feeling that a godot program is not as simple as a SQL or a Python program.  Which only need the text of the program and an interpreter to run it.

DavidCooperWCU:
In the activecode container in a book, the index.html is loaded in an iframe. That uses the other files in the `godot-shell` directory:

```sh
% ls -1 ext/godot-shell
index.apple-touch-icon.png
index.apple-touch-icon.png.import
index.audio.position.worklet.js
index.audio.worklet.js
index.html
index.icon.png
index.icon.png.import
index.js
index.pck
index.png
index.png.import
index.wasm
```

Then, each activecode program lists a .pck file that is loaded from `external/gdpractice`.
The activecode program also specifies the starting scene for where the student's gdscript code is injected.

So, the problems in the book will put the required problem specific .pck files in the ext/gdpractice folder of the book, and we are just working out where to put the `godot-shell` folder so that it gets properly copied into `_static`, and then I'll need to change `activecode_gdscript` to look in `_static` instead of `external`, but that's simple enough.

bnmnetp:
What is in the `pck` files?  Is that something the student modifies?

Is `index.wasm` the gdscript interpreter or is there an other wasm file for that?

What we want to have packaged up by runestone is the infrastructure to run one of these programs.  We want authors to be able to make their programs in PreTeXt markdown, and have PreTeXt + Runestone do the rest.  PreTeXt should get all of the the things specified in PreTeXt translated to html and copied to the right place.  Runestone should just render the editor, provide a run button and be able to call a function that runs the code entered by the author or modified by the student.

If all of the `index.*` files are common to ALL gdscript programs then authors should not have to think about them they just need to be somewhere that the interpreter can load them.  If each example/program needs its own `pck` file how are those going to be authored?


@ascholerChemeketa keeping you in the loop as you may have more experience with Godot than I.

DavidCooperWCU:
Yes, the `index.wasm` file is the godot-shell wasm. Agreed, the author should not need to think about them.

The student will not be making pck files. There is a pck file for the wasm, and then there is an additional pck file for each example program. 

With the current design, the author provides the pck file for each problem. That file has a directory structure that includes testing code and environment for the particular problem. The system works this way so that each problem can have its own assets specific to the problem. Since this is a game, there is either a 2D or 3D environment for the program. This environment gets loaded from the problem specific pck file.

Authoring Godot Practice problems is a little bit more involved because of this.

We may want to settle on some default environments that are provided. This is straightforward. We'd just provide the default environments from rs.  This would require a way for testing code injection in addition to the student code injection. That shouldn't be too hard since there is a place for it in the pretext. It's just different than my initial design.

I was looking at the way the sql.js wasm file was added:

```javascript
new CopyPlugin({
                    patterns: [
                        {
                            // sql.js support: this wasm file will be fetched dynamically when we initialize sql.js. It is important that we do not change its name, and that it is in the same folder as the js.
                            from: "node_modules/sql.js/dist/sql-wasm.wasm",
                            to: ".",
                        },
                    ],
                }),
```

Based on that, I think we could do:

```javascript
new CopyPlugin({
                    patterns: [
                        {
                            // sql.js support: this wasm file will be fetched dynamically when we initialize sql.js. It is important that we do not change its name, and that it is in the same folder as the js.
                            from: "node_modules/sql.js/dist/sql-wasm.wasm",
                            to: ".",
                        },
                        {
                            // godot support: this Godot-script file will be fetched dynamically. It is important that we do not change its name or try to minimize any of the code.
                            from: "extra_dependencies/godot-shell",
                            to: "./godot-shell",
                            // need this: Tells webpack to skip parsing or optimizing these files
                            info: { minimized: true } 
                        },
                    ],
                }),
```

DavidCooperWCU:
I made the change I suggested above, and it works.

Are you okay with the following directory `bases/rsptx/interactives/extra_dependencies`?

Or should I place it somewhere else?

DavidCooperWCU:
> What we want to have packaged up by runestone is the infrastructure to run one of these programs. We want authors to be able to make their programs in PreTeXt markdown, and have PreTeXt + Runestone do the rest. 

I've made some changes so that this can happen for a subset of problems, and I will work on expanding this.

My current plan is to have 2 folders in `bases/rsptx/interactives/extra_dependencies` that each get copied into `_static`.

- `godot-shell` - the wasm running code
- `gdpractice` - the place for .pdk templates usable by instructors for basic functions or a predetermined environment. These will not have default solutions, since the problems will be defined by the instructor.

If you agree to the location of the files and the scripting change in the `webpack.config.js` file to do the copying, then I'll commit and push that change for testing. If you'd prefer a different directory structure, please let me know.

ascholerChemeketa:
The CopyPlugin mechanism would mean the WASM files are bundled with every book, correct? 

Again, that is what I am concerned about. That would bloat the size of the _static directory of every book by a factor of something like 3x. Runestone interactives are the default mechanism for all interactive exercises. A math book that is using Runestone to add Parsons problems should not have to bundle 40Megs of unused wasm that is only needed for books using GDScript.

Even worse, extrapolate this pattern out over a few other new languages/tools.

It really feels like these should be loaded from a server and not bundled directly into the book.

bnmnetp:
@ascholerChemeketa raises a very good point.  One reason why I have stuck with skulpt all these years is that it is **much smaller** than going with one of the wasm implementations of Python in the browser.

From what I can tell there is not a standard place to get the distributed files for Godot.  Which means we would have to self host them on our pseudo CDN on Runestone.academy.  Or perhaps there is something somewhere I don't know about.

At a minimum I think we would need to version things on the CDN so that books are loading the version of Godot the author intends and not just whatever unversioned index.wasm happens to be there.

DavidCooperWCU:
> The CopyPlugin mechanism would mean the WASM files are bundled with every book, correct?
> 
> Again, that is what I am concerned about. That would bloat the size of the _static directory of every book by a factor of something like 3x. Runestone interactives are the default mechanism for all interactive exercises. A math book that is using Runestone to add Parsons problems should not have to bundle 40Megs of unused wasm that is only needed for books using GDScript.
> 
> Even worse, extrapolate this pattern out over a few other new languages/tools.
> 
> It really feels like these should be loaded from a server and not bundled directly into the book.

This is why I had the wasm in the book itself, instead of in rs.

I am not attached to a solution. 

Whatever is easiest to maintain.

Right now the SQL wasm is attached to every book that uses activecode.

bnmnetp:
Authors should not have to even know what a .wasm is in order to write a book.

Yes, we have the sql wasm bundled  but the wasm for sql is just 1 MB.

If you can send me the index.wasm or point me to where I can grab it I will put it in:

https://runestone.academy/cdn/runestone/godot-X.Y/image.wasm

DavidCooperWCU:
> Authors should not have to even know what a .wasm is in order to write a book.

I agree.


Does your pseudo CDN do remapping? If not, then we'll need the whole godot shell directory to be on `runestone.academy/cdn/runestone/godot-shell-4.6.3`

The wasm file is in an [older commit](https://github.com/DavidCooperWCU/pretext/tree/1715d5dc85adbdd7f69a2019fed7ede3a270edb9/examples/sample-book/ext/godot-shell) of my fork of the pretext book in [ext/godot-shell/index.wasm](https://github.com/DavidCooperWCU/pretext/blob/1715d5dc85adbdd7f69a2019fed7ede3a270edb9/examples/sample-book/ext/godot-shell/index.wasm).

If it's all you need, then you can download it from there. If you need the rest of the files also, then I'm happy to zip them up and share them with you.

If the the [corresponding files](https://github.com/DavidCooperWCU/pretext/tree/1715d5dc85adbdd7f69a2019fed7ede3a270edb9/examples/sample-book/ext/godot-shell) that are needed to run with it will not be in the same directory as the wasm file, then should I put those in the directory structure as specified above? 


Similarly, with the pck practice problem files. 

One option would be to have some default files as part of rs, and the other would be to put them in the cdn directory. In that case, I think it could look like this, and there would be no need to do standard CDN remapping of the files:

- https://runestone.academy/cdn/runestone/godot-X.Y/godot-shell  *to hold all of the shell fiiles*
- https://runestone.academy/cdn/runestone/godot-X.Y/gdpractice *to hold the ready to use .pck files*

bnmnetp:
If it expects all those other js and image files to be siblings in the same folder that is fine. They can all go on our cdn.

I think the closest analogy to the pck files I can think of are the .db files that we use for the sql an instructor can supply those in a PreTeXt external folder - Can you configure gdscript to look for the pck files in a particular folder that isn't the same as the wasm folder?

DavidCooperWCU:
> Can you configure gdscript to look for the pck files in a particular folder that isn't the same as the wasm folder?

Yes, That's how it's configured now. The index.pck folder should be with the .wasm file, but the individual practice .pck files are in the external directory.

The only reason we might want to have some .pck files on your cdn in a different directory is so that instructors could have default environments that they could make problems from. Perhaps one without graphics, just to test functions, one with a sprite to test moving the sprite programmatically, one with 2 objects to test collisions, etc. It's not necessary, but it would lower the barrier to entry for authors who don't want to build their own test environments for their problems.

DavidCooperWCU:
> If you can send me the index.wasm or point me to where I can grab it I will put it in:
> 
> https://runestone.academy/cdn/runestone/godot-X.Y/image.wasm

Since we'll probably want a record of the versions of the godot-shell, are you all right with me making an `extra_dependencies` folder in `interactives` without the copy script? Then I would put `godot-shell` in for the cdn, and `gdpractice` for the default environment .pck files.

I'm also keeping the `godot-shell` in my [pretext conversion of the godot documentation](https://github.com/WCU-CS-CooperLab/godot-docs-pretext) under the [`godot-shell-export`](https://github.com/WCU-CS-CooperLab/godot-docs-pretext/tree/master/gdscript-shell-export). So, if you just want to take the files from their as they are updated, that could work as well.

Please, let me know what you'd prefer, and also let me know the full link of the `godot-shell` once it is copied so I can update the link in the code accordingly.

bnmnetp:
Yes, make an extra_dependencies folder and under that have a versioned godot-shell that I can just copy up to the cdn?  That way you have control over exactly the version and what I copy up to the cdn.

DavidCooperWCU:
>Yes, make an extra_dependencies folder and under that have a versioned godot-shell that I can just copy up to the cdn? That way you have control over exactly the version and what I copy up to the cdn.

Sounds good, Thanks!

 I just added it. Once you have it in place I'll change the links and do some more testing.

bnmnetp:
OK, use this url:  `https://runestone.academy/cdn/runestone/godot-shell-4.6.3`

DavidCooperWCU:
This should be all set now. I had to make a few changes to the index.pck and index.js, so you’ll need to copy those over to the cdn.

Thanks!

DavidCooperWCU:
The current code works with [this pretext branch](https://github.com/DavidCooperWCU/pretext/tree/master) which has  [this open PR](https://github.com/PreTeXtBook/pretext/pull/2892) that is waiting on this PR before it can be completed.

bases/rsptx/interactives/runestone/activecode/js/activecode.js
```diff
@@ -68,6 +72,8 @@ const codeIndent = () => Number(window.getComputedStyle(document.body).getProper
 export class ActiveCode extends RunestoneBase {
     constructor(opts) {
         super(opts);
+        // for gdscript mode in codemirror
+        registerGDScriptMode(CodeMirror);
```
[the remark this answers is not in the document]
DavidCooperWCU: How do the other modes get registered? Would it be better to move the code to there?

bases/rsptx/interactives/runestone/activecode/js/activecode_gdscript.js
```diff
@@ -0,0 +1,420 @@
...
+    // -------------------------------------------------------------------------
+    _sendToShell(payload) {
+        try {
+            // Specify your remote server origin for security
+            const remoteOrigin = "https://runestone.academy"; 
+            
+            this.godotIframe.contentWindow.postMessage({
+                type: "loadExercise",
+                payload: payload
+            }, remoteOrigin);
```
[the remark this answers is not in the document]
DavidCooperWCU: this.shellUrl should be an absoluteUrl, but if it's not, then using the window.location.href probably makes sense, so this is probably the right change.

bases/rsptx/interactives/runestone/activecode/js/activecode_gdscript.js
```diff
@@ -0,0 +1,420 @@
...
+    //   2. data.source === "godot-activecode" — ensures we only handle
+    //      messages intentionally sent by Shell.gd, ignoring anything else
+    //      posted to the window (browser extensions, other widgets, etc.).
+    // -------------------------------------------------------------------------
+    _listenForMessages() {
+        window.addEventListener("message", (event) => {
+            // Guard 1: message must come from this instance's iframe.
+            if (event.source !== this.godotIframe.contentWindow) return;
+
+            var data = event.data;
```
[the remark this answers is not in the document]
DavidCooperWCU: this probably makes sense.

DavidCooperWCU:
> **bases/rsptx/interactives/runestone/activecode/js/acfactory.js:15**
> 
> * `TimedGodotActiveCode` is used below but is never imported from `./timed_activecode`, which will throw a `ReferenceError` when creating timed GDScript ActiveCode instances. Also, this file’s other local imports include the `.js` extension; importing `./activecode_gdscript` without it is inconsistent and can break in stricter ESM/bundler setups.
> 
> ```
> import GodotActiveCode from "./activecode_gdscript";
> 
> 
> import {
>     TimedActiveCode,
>     TimedLiveCode,
>     TimedJSActiveCode,
>     TimedHTMLActiveCode,
>     TimedSQLActiveCode,
> } from "./timed_activecode";
> ```

This probably should be fixed as well. adding TimedGodotActiveCode to the part above.

DavidCooperWCU:
@bnmnetp would you like me to apply copilot's suggestions on my branch and test them out. If I accept the changes here do they get committed to my fork or do they stay in the pull request somehow?

bnmnetp:
I would guess they go on your branch because there wouldn't be anyplace for them to go on main.  So, I would say apply them, unless you disagree with one or more, then comment as to why so I can resolve it.   Then test them locally.

DavidCooperWCU:
I made the changes and they worked.

DavidCooperWCU:
@bnmnetp I applied the changes and made one additional update to the godot shell so that split screen testing can be done. I'm not sure if you want to have copilot do another review, or if you just want to review it yourself.

Please, make sure to copy the updated index.js and index.pck files to https://runestone.academy/cdn/runestone/godot-shell-4.6.3

Thanks!

bases/rsptx/interactives/runestone/activecode/js/activecode.js
```diff
@@ -68,6 +72,12 @@ const codeIndent = () => Number(window.getComputedStyle(document.body).getProper
 export class ActiveCode extends RunestoneBase {
     constructor(opts) {
         super(opts);
+        // for gdscript mode in codemirror (register once)
```
bnmnetp: having this right after the call to super makes it seem very important.  It would be better to move it somewhere where it was only called when necessary.  Even moving it to the GodotActiveCode constructor seems much better.

bnmnetp (changes requested):

bases/rsptx/interactives/runestone/activecode/js/activecode.js
DavidCooperWCU: I tried moving it to the GodotActiveCode constructor, and it made it so that the first gdscript activecode block isn't highlighted.

Other code mirror imports don't require a function call, so maybe there is a way to change the function definition so that the import works without the registration call.

If you think it's really important to change this, then I'll look into it.

bnmnetp: I try to keep the bundles of things that need to get fetched and loaded in the browser as small as possible.  This setup means that every page of every book that has an activecode will now need to import the gdscript stuff which is unnecessary for all books except yours.  Seems like there should be a way around this.

DavidCooperWCU: What if I put it in acfactory, and only load it if gdscript is the language?

DavidCooperWCU: Let me know if you like this change, or you'd like something different.

bnmnetp:
OK, I've copied everything up to the cdn again.

DavidCooperWCU:
Thanks. It’s working with the new pretext changes!

DavidCooperWCU:
This is now associated with [pretext pr 3024](https://github.com/PreTeXtBook/pretext/pull/3024). I am closing the other pretext pr because it got too messy.

DavidCooperWCU:
@bnmnetp now that the pretext pr has been merged, this should be ready except for one question.

The one question that I have, is whether we add a couple of localhost ports in the pseudo-cdn for testing or not. Right now godot will only run if the book is hosted on runestone or the same location as the cdn. When I was testing I used my own pseudo-cdn that allowed localhost as well.

Please let me know if you'd like me to add that. 

These are the additional entries that I needed in the `index.js` file for testing.
```
         "https://localhost:8129",
         "http://localhost:8130",
         "https://localhost:8130",
```

bnmnetp:
What do you mean by pseudo-cdn?

DavidCooperWCU:
Maybe it's an actual cdn? I mean the place you are hosting the wasm files.

bnmnetp:
That server is locked down and lives behind cloudflare.  I don't want to open any ports there for any other reason.

What do you need localhost:8129 and localhost:8130 for ??

DavidCooperWCU:
You won't need to open any ports. The godot.wasm is locked down to only allow runestone.academy and the server where the book is served from. 

The localhost entries are so that you can run a pretext book locally and see the godot activecode working.

DavidCooperWCU:
Also, the change isn't necessary, authors could use the version that I use for testing. 

It would just require them to set an extra string parameter.

bnmnetp:
I think we should stick with the stringparam for now.  If author authors want to use GDScript then we can see.

Localhost stuff gets complicated when you are building docker containers, so I'm not even sure which localhost you want to allow people to connect to or if those are new docker containers?  Also the author interface would need to support these localhost interfaces if you want to do that and be able to build books from the author interface.  If a single string parameter lets you handle your book then that is a good starting point.  If others want to start writing Godot then we'll need to figure out how to support that.  

Have you added a Godot example to the pretext sample book?  That is important as I use that a lot for quickly making sure that we don't have regressions in existing behavior when changing the code.

DavidCooperWCU:
Sounds good.

There are 3 activecode examples in the pretext sample book.

PR #1238, opened 2026-06-16
iss-1214: source and answer area height fix.

loua:


bnmnetp:
This definitely fixes the issue, but it also leaves a number of the examples with a lot of extra space at the bottom of the source box.  Its definitely better than not having enough room.

But it leaves room for further improvement to make the source box height match the amount of needed space more closely.

bnmnetp:
merged

loua:
@bnmnetp I am looking at this again, and have generated a set of parsons activities to test various scenarios. 

I was able to reduce the empty space at the bottom, but noticed that there are some layout issues with math blocks. Interestingly, the layout corrects when clicking the Reset button. Here are before (initial rendering on page load) and after reset screenshots.

**Before Reset**
math-before-reset
**After Reset**
math-after-reset

I'll continue to look at this issue. I've attached the test activities I am using. Let me know if you have additional examples beyond these that I can test.
[parsons-overflow-tests-v2.xml](https://github.com/user-attachments/files/29137665/parsons-overflow-tests-v2.xml)
[parsons-overflow-tests-v1.xml](https://github.com/user-attachments/files/29137667/parsons-overflow-tests-v1.xml)

PR #1240, opened 2026-06-16
Adds spaced practice to the new gradebook

sethbern:
Adds spaced practice grading to the instructor gradebook. This update calculates practice scores separately from normal assignments and conditionally adds a Practice column only when spaced practice is configured for the course.


bases/rsptx/assignment_server_api/routers/instructor.py
```diff
@@ -333,6 +333,84 @@ async def get_assignment_gb(
...
+                ps.max_practice_questions or 0
+            )
+
+            for _, prow in practice_counts.iterrows():
+                points_received = float(ps.question_points or 0) * float(
+                    prow.practice_completion_count or 0
+                )
+                practice_by_user_id[prow.user_id] = "{0:.2f}".format(
+                    100 * points_received / total_possible_points
+                )
```
bnmnetp, line 435: There is a course attribute that controls whether we display points or percentages in the gradebook.  Make sure this obeys that setting.

sethbern:
@bnmnetp this should be good to merge now

PR #1241, opened 2026-06-16
Add A/B experiment for synchronous PI (verbal vs. text chat)

sethbern:
This adds in an A/B experiment for synchronous Peer Instruction. After vote 1, verbal discussion groups get split into two conditions: half keep discussing verbally (condition 0) and half switch to the text chat (condition 1). 

A few things about how the splitting works:
- The source of truth is the **verbal** `peergroup` events (the "who did you talk to?" submissions after vote 1).
- Verbal groups are never split across conditions since they're sitting together physically, the whole group goes to one condition or the other.
- Clusters get shuffled and assigned with greedy least-loaded placement so the split stays close to a 50/50 split.

**This is off by default and has to be turned on per course**, so it won't touch any existing courses. It can be enabled by noting it in the db using the `enable_ab` course attribute:
```sql
INSERT INTO course_attributes (course_id, attr, value)
VALUES (<course_id>, 'enable_ab', 'true');
```

If possible could the course test_py4e-int_api have the A/B test feature turned on once this is merged?

Once that's set, a "Run A/B Experiment" button shows up in the Discuss step on the instructor dashboard. Courses without it see the normal PI flow.

Also fixed a couple things from testing/feedback while I was in here:
- Instructor's own answer was getting counted as a student response
- Text chat now shows a "Text Chat Ready" message instead of the stale "wait for your instructor" banner when it starts.
- You can no longer select yourself in the verbal group dropdown.


bnmnetp:
FYI:  `rsmanage addattribute` will allow you to create the needed course attribute without resorting to a database query.

PR #1243, opened 2026-06-18
f-1220 Huge summer Update(styles, grading, tests)

morozov-av:
New UI foundation                                                                                                                                                                                                                                                                                                                                                                                 
  - Migrated the entire UI off PrimeReact (primereact / primeflex / primeicons) - its theming and component quirks cost more than they saved                                                           
  - Rebuilt every screen on Mantine 7          
  - Applied one consistent color scheme via CSS Modules + shared --rs-* tokens, replacing scattered hardcoded values. Consistent hover / focus-visible / disabled / selected states.   

Expanded grader                                                                                                                                                                                                                                                                                                                                                                                        
  - Batch re-grade with dry-run preview (/regrade/preview) returning a before/after diff, then commit (/regrade) with options for deadline enforcement, manual-grade protection, and total recomputation.
  - New endpoints: threshold grading, manual total override, total recompute, grade release, gradebook data + CSV export, per-question answer history (brute-force detection) and answer stats.          
  
Tests                                                                                                                                                                                                                                                                                                                                                                                                       
  - ~185 Vitest unit specs: utils, preview generators, questionJson, sanitize, date/object/json, HTML regeneration.                                                                                      
  - Full Playwright e2e suite: every editor, wizard, browse/search, readings, complete grader cycle (gradebook, threshold, manual total, release, roster security)                                                                                           
  - Backend pytest for grader routes and threshold logic.


morozov-av:
image
image
image

bnmnetp:
Wow!!

That is one huge update.  I've built it and played around a bit and it looks quite good.  I think I will merge today to get it out there.  Plus we are doing an online workshop the week of July 6 so it is great to have this available for that.

Lets plan to meet next week.  To talk about whats next and some feedback from the Proteus workshop this week including:

1. support Donet questions in the grader
2. support Webwork questions in the grader
3. Add support for hiding student names in the grader
4. Using a consistent look/feel across all instructor page nav bars
5. Using a consistent look/feel across all student page nav bars
6. How do I run the unit tests? Should we add them to CI? and build test
7. Teachers think that the average number of attempts would be more useful than the fully correct column

I'll try to get these added to the project before we meet as well.

bnmnetp (commented):
Review focused on the Python changes (grading / regrade), as requested. Three issues, each with a suggested fix inline. I'd treat **#1 as a blocker** (silent loss of instructor-entered grades); #2 and #3 are fix-before-merge. I did not deep-review the React/Playwright changes.

components/rsptx/grading_helpers/regrade.py
```diff
@@ -0,0 +1,419 @@
...
+    try:
+        existing = await fetch_question_grade(sid, course.course_name, div_id)
+        if existing is not None and existing.score is not None:
+            item.old_score = float(existing.score)
+        if (
+            existing is not None
+            and existing.comment
+            and existing.comment != MANUAL_COMMENT
+            and not options.overwrite_manual
+        ):
```
bnmnetp, line 194: **Bug (data loss): manual grades with an empty comment are not protected.**

The manual-grade write path (`POST /grade`) stores `row.comment = payload.comment or ""`, so an instructor who grades a question without typing a comment ends up with `comment == ""`. That is falsy, so the `and existing.comment` clause short-circuits and this row is **not** treated as manual — a default regrade (`overwrite_manual=False`) silently overwrites the instructor's score and restamps it `"autograded"`.

Since the autograder always stamps `comment="autograded"`, any row whose comment is *not* `"autograded"` (including `""`) came from a human. Drop the truthiness clause and gate on a real score instead:

```suggestion
        if (
            existing is not None
            and existing.score is not None
            and existing.comment != MANUAL_COMMENT
            and not options.overwrite_manual
        ):
```

bases/rsptx/assignment_server_api/routers/grader.py
```diff
@@ -620,3 +634,358 @@ async def upsert_grade(
...
+    assignment = await fetch_one_assignment(payload.assignment_id)
+    if not assignment or assignment.course != course.id:
+        return make_json_response(
+            status=status.HTTP_404_NOT_FOUND, detail="Assignment not found"
+        )
+
+    instructor_ids = {
+        u.username for u in await fetch_course_instructors(course.course_name)
+    }
+    sids = [s for s in payload.sids if s not in instructor_ids]
```
bnmnetp, line 761: **Bug: an all-instructor `sids` list silently recomputes the whole course.**

`recompute_totals_for` treats an empty `sids` as "recompute every student" (see `regrade.py:359-362`). If the client sends a non-empty `payload.sids` that filters down to empty here (e.g. it contained only instructors), `sids` becomes `[]` and *every* student in the course is recomputed and re-pushed to the LMS — the opposite of the requested scope. Guard the filtered-to-empty case:

```suggestion
    sids = [s for s in payload.sids if s not in instructor_ids]
    if payload.sids and not sids:
        return make_json_response(
            status=status.HTTP_200_OK,
            detail={"assignment_id": assignment.id, "students": 0},
        )
```

components/rsptx/grading_helpers/regrade.py
```diff
@@ -0,0 +1,419 @@
...
+    students. When ``sids`` is empty/None every student in the course is
+    recomputed. Returns the number of students processed.
+
+    This is used by the manual multi-grade flow, where individual grades are
+    written through ``POST /grade`` (which does not itself recompute totals).
+    """
+    users = await fetch_users_for_course(course.course_name)
+    user_map: Dict[str, AuthUserValidator] = {u.username: u for u in users}
+
+    if sids:
```
bnmnetp, line 359: Related to the `/recompute_totals` finding: this `if sids:` overloads `[]` and `None` to both mean "all students", so a caller passing an explicitly-filtered (but now-empty) list can accidentally trigger a full-course recompute. Consider distinguishing `None` (all) from `[]` (none) here; the endpoint guard suggested in `grader.py` is the minimal fix.

components/rsptx/build_tools/build.py
```diff
@@ -981,8 +981,8 @@ def test(config):
         console.print("Skipping tests and format checks", style="bold yellow")
         return
     checks = [
-        ["black", "--check", "--config", "pyproject.toml", "bases"],
-        ["black", "--check", "--config", "pyproject.toml", "components"],
+        ["black", "--config", "pyproject.toml", "bases"],
+        ["black", "--config", "pyproject.toml", "components"],
```
bnmnetp: **Regression: `build test` now reformats instead of checking.**

Dropping `--check` means black rewrites files in place during the test/CI step, so it can no longer fail on misformatted code and it mutates the working tree mid-run. If auto-format is desired, put it in a separate command; the check step should verify:

```suggestion
        ["black", "--check", "--config", "pyproject.toml", "bases"],
        ["black", "--check", "--config", "pyproject.toml", "components"],
```

morozov-av: I did this intentionally, since I got tired of running it manually on every commit :) But I agree with the comment. What tool do you use for such reviews? Maybe I could run it before creating the  PR (if it's available in my country). I'll look at the rest of the questions tomorrow and fix them.

PR #1245, opened 2026-06-22
Add Likert reflection to async PI (opt-in per course) with logging

sethbern:
Added an opt in Likert scale step to the peer instruction flow that appears after students submit their second vote and before they move to the next question. The next question button remains locked until the Likert responses are completed, and the form resets for each new question.

Also expanded the logging to capture additional peer instruction data, including the async mode type, selected theme for personalized LLM interactions, and students’ Likert scale responses.

The Likert step is off by default and has to be enabled per course, so it won't affect any existing courses. It's gated behind the enable_likert course attribute, which can be set with:

`rsmanage addattribute --course <course_name> --attr enable_likert --value true`

Could the course `async-pi-study` have this enabled once it's merged?

### DIFF
--- a/bases/rsptx/assignment_server_api/routers/peer.py
+++ b/bases/rsptx/assignment_server_api/routers/peer.py
@@ -550,6 +550,7 @@ async def get_peer_async(
 
     course_attrs = await fetch_all_course_attributes(course.id)
     latex_macros = course_attrs.get("latex_macros", "")
+    enable_likert = course_attrs.get("enable_likert", "false") == "true"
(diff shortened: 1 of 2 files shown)


sethbern:
@bnmnetp I addressed the Copilot comments. Once it's merged, could you enable it for async-pi-study and test_py4e-int_api?

rsmanage addattribute --course async-pi-study --attr enable_likert --value true
rsmanage addattribute --course test_py4e-int_api --attr enable_likert --value true

PR #1249, opened 2026-06-23
f-1221 Add JSON import functionality for exercise creation and validation

morozov-av:
https://github.com/user-attachments/assets/a545aa12-32d8-475e-927d-23f734cdcdfa

### DIFF
--- a/bases/rsptx/assignment_server_api/assignment_builder/src/components/routes/AssignmentBuilder/components/exercises/components/CreateExercise/CreateExercise.spec.tsx
+++ b/bases/rsptx/assignment_server_api/assignment_builder/src/components/routes/AssignmentBuilder/components/exercises/components/CreateExercise/CreateExercise.spec.tsx
@@ -15,8 +15,23 @@ vi.mock("@components/routes/AssignmentBuilder/hooks/useAssignmentRouting", () =>
   })
 }));
 
+vi.mock("react-redux", () => ({
+  useSelector: () => [{ value: "python", label: "Python" }]
+}));
+
 vi.mock("./components/ExerciseFactory", () => ({
-  ExerciseFactory: ({ type }: { type: string }) => <div data-testid="factory">{type}</div>
+  ExerciseFactory: ({
+    type,
+    initialData
+  }: {
+    type: string;
+    initialData?: { statement?: string };
+  }) => (
+    <div data-testid="factory">
+      {type}
+      <span data-testid="factory-statement">{initialData?.statement ?? ""}</span>
+    </div>
+  )
 }));
 
 vi.mock("./components/ExerciseTypeSelect", () => ({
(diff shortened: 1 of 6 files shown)


bnmnetp:
## Code review — JSON import for exercise creation

Reviewed the diff (`main...f-1221`). The parse/validate utility (`importQuestionJson.ts`) and the modal are solid and well-tested. The issues are in the `CreateExercise.tsx` integration.

### 1. (High) Stale `importedData` leaks into the next created exercise — wrong `question_type` gets saved

`effectiveInitialData = importedData ?? initialData` (`CreateExercise.tsx:114`), but `importedData` is only cleared on `resetForm` (line 60). Neither `handleTypeSelect` (65–72) nor `handleCancel` (74–81) clears it, and the in-form Cancel returns to the type-selection view **without unmounting** `CreateExercise` (it calls `setSelectedType(null)`, not the parent `onCancel`).

Repro: Paste JSON for `mchoice` → land in the factory → click Cancel → pick a *different* type (e.g. `activecode`). `ExerciseFactory` remounts seeded with the leftover `importedData`, which still carries `question_type: "mchoice"` (set at line 103) plus the old field values. `useBaseExercise` merges `{...getDefaultFormData(), ...initialData}`, so the new activecode form inherits `question_type: "mchoice"` and the stale statement, and saves a corrupted exercise.

**Fix:** clear `importedData` (and bump `factoryKey`) in `handleTypeSelect` when the user picks a type manually.

### 2. (Medium) "View / Replace JSON" shows stale JSON after a replace, and re-applying discards the replacement

`currentJson` is memoized on `[isEdit, initialData]` (83–96) and passed to the modal as `initialJson` (line 178). After a user replaces the JSON via import, `importedData` updates but the `initialData` prop does not, so `currentJson` keeps reflecting the *original* question. The modal resets its textarea to `initialJson` on open, so reopening "View / Replace JSON" shows the pre-import JSON. Clicking Apply then re-imports the original and silently throws away the replacement (in-form edits are never reflected either).

### 3. (Low) `handleImportApply` can throw on empty `languageOptions`

`handleImportApply` calls `mergeQuestionJsonWithDefaults(languageOptions, data)` (line 99) → `getDefaultQuestionJson` does `languageOptions[0].value`, which throws if `languageOptions` is an empty array (dataset store not yet populated). The import button newly exposes this fragility on an interactive click path. Pre-existing pattern (shared with `EditView`), but worth a guard at the source.

---

**Minor / non-blocking:** the `"object"` branch in `matchesType` (`importQuestionJson.ts`) is dead — no rule uses type `"object"`. Validation is intentionally shallow (top-level presence + type only), which the tests document; malformed array elements (e.g. `optionList: [1,2,3]`) pass and rely on downstream components tolerating them.

bnmnetp:
### Update: all three findings fixed

- **#1 (High) — stale `importedData` leak:** `handleTypeSelect` now clears `importedData` and bumps `factoryKey`, so picking a type manually after an import starts from defaults (no leftover `question_type`/fields). Regression test added.
- **#2 (Medium) — stale "View / Replace JSON":** `currentJson` is now derived from `effectiveInitialData` (`importedData ?? initialData`) instead of the original prop, so the modal reflects an in-session import/replace rather than the original exercise. Regression test added.
- **#3 (Low) — empty `languageOptions` crash:** `getDefaultQuestionJson` now uses `languageOptions?.[0]?.value` instead of `languageOptions && languageOptions[0].value` (the old guard let an empty array through). Fixed at the source, so `EditView` and the import path both benefit.

All affected suites green (159/159), with new regression tests confirmed to fail on the pre-fix code.

PR #1258, opened 2026-06-25
Small fixes for A/B testing from feedback

sethbern:
I tested this with the peer instruction grant group and everything worked as expected. The only change is that students who didn't vote still stay in the same group as the people they talked to. Before, they could accidentally get split up which would not be the desired outcome for the experiment.

### DIFF
--- a/bases/rsptx/assignment_server_api/routers/peer.py
+++ b/bases/rsptx/assignment_server_api/routers/peer.py
@@ -716,11 +716,12 @@ async def make_pairs(
             for p in answerers:
                 if p in clustered:
                     continue
-                grp = {
-                    s
-                    for s in find_set_containing_string(in_person_groups, p)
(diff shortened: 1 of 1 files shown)


bnmnetp:
The grammar fixes are an easy accept I think.  The only one I have curiosity about is the double call to `grp.add(p )` which just seems like an easy fix also?

sethbern:
Yeah I agree these are all easy accepts.

bnmnetp:
Thanks.  Will merge now.

The only improvement Claude recommended was to refactor the make_pairs endpoint to use a helper function to do the clustering so that we could write unit tests against it.  That may be something to do if we have further issues or the next time we need to refine the clustering.

bnmnetp:
Or maybe I'll have claude do that and see what it looks like...

PR #1260, opened 2026-06-26
Extract A/B verbal-cluster splitting into a testable helper

bnmnetp:
## Summary

Follow-up to the `ab-fixes` merge. The A/B verbal-cluster building and
condition-balancing logic was inline inside the `make_pairs` request handler,
which made it impossible to unit test (it needs a DB, auth, and a live request).
This PR extracts that logic into a pure `split_ab_conditions()` function and
adds unit tests.

**No behavior change** in `make_pairs` — the inline block is replaced by a call
to the new helper.

## What changed

- `split_ab_conditions(answerers, in_person_groups, rng=random)` returns the
  `(peeps_in_person, peeps_in_chat)` partition. It contains the cluster-building
  (`find_set_containing_string` + `grp -= clustered`), greedy ~50/50 balancing,
  and the "seed each condition when possible" adjustment.
- Randomness is injectable via the `rng` parameter so tests can pass a seeded
  `random.Random` for deterministic results.
- `make_pairs` now calls the helper instead of inlining ~55 lines.

## Tests

New `test/bases/rsptx/assignment_server_api/test_make_pairs.py` (9 tests,
pure/sync, no DB) covering:

- singletons (no verbal partner) go to chat
- a non-voting recorded partner stays in the same condition as the voter
- recorded verbal groups are never split across conditions
- overlapping recorded groups stay disjoint with no duplicates (the
  `grp -= clustered` guard)
- both conditions are populated when possible
- a lone multi-person cluster seeds the verbal side
- greedy placement keeps the two conditions roughly balanced
- determinism with a seeded rng

All 9 pass; `black`/`flake8` clean on the changed files; assignment service
rebuild (`uv run build -s assignment dev`) passes including the full suite.

### DIFF
--- a/bases/rsptx/assignment_server_api/routers/peer.py
+++ b/bases/rsptx/assignment_server_api/routers/peer.py
@@ -627,6 +627,92 @@ async def get_peer_async(
     return templates.TemplateResponse("assignment/student/peer_async.html", context)
 
 
+def split_ab_conditions(
+    answerers: list[str],
+    in_person_groups: list[set[str]],
+    rng: random.Random = random,
+) -> tuple[list[str], list[str]]:
+    """Split answering students into the verbal (in-person) and chat conditions
+    for an A/B peer-instruction experiment.
+
+    Students recorded in the same verbal-discussion group share a single
+    condition so that verbal partners are never split into text chat. The full
+    recorded group is kept, including partners who did not vote on this question,
+    so a non-voting partner is still assigned the same condition.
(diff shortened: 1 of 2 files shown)


bnmnetp:
@sethbern - Any comments on this?  See also #1261 which claude flagged as an edge case worth thinking about.

sethbern:
Looks good to me. I can take a look at at fixing #1261.

PR #1262, opened 2026-06-26
Fix transitive verbal cluster split (fixes #1261)

### LINKED ISSUE #1261: A/B peer pairing: transitively-linked verbal groups can be split across conditions
## Background

`split_ab_conditions()` in `bases/rsptx/assignment_server_api/routers/peer.py`
(extracted in #1260) assigns students to the in-person vs. text-chat conditions
of the A/B peer-instruction experiment. The core invariant it tries to uphold is
that **students who discussed verbally are kept in the same condition** so they
are never split into text chat.

Clusters are built by seeding from each answerer `p` and unioning every recorded
verbal group that *directly* contains `p`:

```python
grp = set(find_set_containing_string(in_person_groups, p))
grp -= clustered      # keep clusters disjoint
grp.add(p)
clustered |= grp
clusters.append(sorted(grp))
```

## The problem

`find_set_containing_string` only unions the recorded sets that directly contain
the seed. It does **not** follow transitive links. When two recorded groups
overlap but no single recorded set contains all members, partners who really are
verbally connected can land in different conditions.

### Example

Recorded groups: `{A, B}` and `{B, C}` (e.g. the seating/peergroup was recorded
slightly differently across two events, but A, B, C were all at the same table).

Processing answerers in order `[A, B, C]`:

1. `A` seeds `find_set(A)` = `{A, B}` → cluster `{A, B}`, `clustered = {A, B}`
2. `B` is already in `clustered` → skipped
3. `C` seeds `find_set(C)` = `{B, C}`, minus `clustered {A, B}` = `{C}` →
   singleton cluster `{C}`

Result: `{A, B}` may go to the verbal condition while `C` (a singleton) is forced
to text chat — even though B and C are recorded verbal partners. The two clusters
can be assigned to **different conditions**.

### Note on the `grp -= clustered` guard

The `grp -= clustered` subtraction added in the `ab-fixes` work is correct and
should stay — it prevents a student from being duplicated across clusters/
conditions (the prior code would list `B` in two clusters). The fragmentation
described here is a *separate, pre-existing* limitation of the per-seed union
approach; the guard does not cause it, and removing the guard would reintroduce
the duplicate-assignment bug. This is an edge case that only triggers when
recorded groups overlap without a single set spanning all members.

## Proposed fix

Compute **connected components** over `in_person_groups` (union-find / disjoint-set,
or a graph component pass) so transitively-linked students form a single cluster
before the balancing step. Concretely: treat each recorded group as edges among
its members, then each connected component becomes one cluster. This makes
clustering order-independent and guarantees verbally-connected students always
share a condition, while preserving disjointness.

## Testing

`test/bases/rsptx/assignment_server_api/test_make_pairs.py` already exercises
`split_ab_conditions` deterministically (seeded `random.Random`). The current
`test_overlapping_groups_stay_disjoint` only asserts that the directly-seeded
pair (A, B) shares a condition; a fix here should add a case asserting that
**transitively** linked students (A–B, B–C) all land in the same condition.

## References

- PR #1260 — extracted `split_ab_conditions` into a testable helper (no behavior
  change); surfaced during its code review.
- Related prior work: the `ab-fixes` branch that fixed the condition-recording
  asymmetry and added the `grp -= clustered` disjointness guard.

## Priority

Low. Requires overlapping-but-not-identical recorded verbal groups, which is
uncommon. Worth fixing for experiment correctness, not urgent.

sethbern:
This fixes issue #1261. It replaces the per-seed union approach with a union-find (connected components) pass so transitively-linked verbal groups.

### DIFF
--- a/bases/rsptx/assignment_server_api/routers/peer.py
+++ b/bases/rsptx/assignment_server_api/routers/peer.py
@@ -651,33 +651,40 @@ def split_ab_conditions(
         ids assigned to the verbal and chat conditions respectively.
     """
 
-    def find_set_containing_string(
-        list_of_sets: list[set[str]], target: str
-    ) -> set[str]:
-        result: set[str] = set()
-        for s in list_of_sets:
-            if target in s:
-                result |= s
-        return result
+    # Build connected components over all recorded verbal groups using union-find
+    # so that transitively-linked students (e.g. {A,B} and {B,C} recorded
+    # separately) always land in the same cluster. A per-seed union approach
+    # would make C a singleton when A is seeded first (issue #1261).
+    parent: dict[str, str] = {}
 
-    # Group the answering students into their recorded verbal-discussion
-    # clusters. A cluster is assigned to a condition as a whole and is never
-    # split, because verbal discussion depends on physical seating.
-    clusters: list[list[str]] = []
-    clustered: set[str] = set()
+    def _find(x: str) -> str:
+        while parent.setdefault(x, x) != x:
+            parent[x] = parent[parent[x]]
+            x = parent[x]
+        return x
+
+    def _union(a: str, b: str) -> None:
+        parent[_find(a)] = _find(b)
+
+    for grp in in_person_groups:
+        members = list(grp)
+        for m in members[1:]:
+            _union(members[0], m)
+
+    # Group answerers into their connected component. Non-voting recorded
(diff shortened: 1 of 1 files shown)


bnmnetp:
Did you do this against the PR with the function and unit tests?

sethbern:
I rebased onto main so the fix is now inside split_ab_conditions.

PR #1264, opened 2026-06-29
Fix/1259 pi websocket state

### LINKED ISSUE #1259: Runestone Issue: Peer Instruction - student occasionally doesn't see partner picker or chat panel
During a live PI session, occasionally a student won't see the partner picker or chat panel appear when the instructor enables it. Instructors noted that this happened in web2py too so it's not new.

This can happen occasionally even if the student never refreshes. My inclination is that any brief WebSocket drop will trigger an automatic reconnect, and there is no catch-up on reconnect so the broadcast is permanently missed. 

My inkling is also that the fix would be storing the current phase in Redis and having the client fetch it on reconnect. I am not sure if this is worth a big change since it seems to be rare, but I am happy to investigate/take on fixing if it is.

sethbern:
Fixes #1259.

This stores the current PI phase in Redis on each phase change and exposes it via `/peer/api/current_state`. On websocket connect/reconnect the client fetches and re-applies the phase, so students who briefly disconnect see the correct UI. A `_catchup` flag prevents side effects (no double vote counts, no `enableNext` navigation).

This also works if a student reloads the page, though I think they should still be discouraged to do that.

I was not sure the best way to test since I cant organically recreate the issue but I tested locally by using `ws.close()` in the console broweser and a full page reload mid-session. If you have any suggestions on testing it I would be open to trying @bnmnetp.

### DIFF
--- a/bases/rsptx/assignment_server_api/routers/peer.py
+++ b/bases/rsptx/assignment_server_api/routers/peer.py
@@ -222,6 +222,7 @@ async def get_peer_dashboard(
             "course_name": course.course_name,
         }
         r.publish("peermessages", json.dumps(mess))
+        r.hset(f"{course.course_name}_state", "current_phase", json.dumps(mess))
     except Exception as e:
         rslogger.error(f"Error initializing Redis: {e}")
 
@@ -883,6 +884,15 @@ async def make_pairs(
             }
             r.publish("peermessages", json.dumps(mess))
 
+        # Store current phase so reconnecting students can catch up (see /api/current_state).
+        # Partner data is per-student and is in partnerdb
+        phase_key = "enableFaceChat" if (is_ab and peeps_in_person) else "enableChat"
+        r.hset(
+            f"{course.course_name}_state",
+            "current_phase",
+            json.dumps({"type": "control", "message": phase_key, "broadcast": False, "course_name": course.course_name}),
+        )
+
         # If doing A/B, also send face-chat groups based on in-person groups
         if is_ab and peeps_in_person:
             # Build a map of username -> full name for this course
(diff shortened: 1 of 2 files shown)


bnmnetp:
A couple of drastic things to try:

1. turn your wifi off and back on again.
2. `docker compose restart book` to restart the book server.

sethbern:
I tested docker compose restart book and the catch-up works correctly on reconnect. I could not test wifi off/on locally since the app runs on localhost, but the catch-up logic triggers on any ws.onopen event regardless of disconnect reason, so it should work there as well.

bnmnetp:
I haven't studied your code, but the question I had was how are you distinguishing between a catch up and a legit new session?

Also, do we need to be concerned about Redis hygiene and have a mechanism to clean up old outdated state?

bnmnetp:
PS - I should have said this right up front, but this is great to see!  Thanks.

bases/rsptx/assignment_server_api/routers/peer.py
```diff
@@ -222,6 +223,7 @@ async def get_peer_dashboard(
             "course_name": course.course_name,
         }
         r.publish("peermessages", json.dumps(mess))
+        r.hset(f"{course.course_name}_state", "current_phase", json.dumps(mess))
```
bnmnetp, line 226: To my earlier comment, maybe the key should use the assignment_id - would not even need the course since assignment id column is unique?

bnmnetp (commented):
Maybe the assignment_id isn't important, but maybe when the instructor starts a new PI session it should initialize the state?  This limits us to one PI session per course, which is already a hard limitation, so not a critique of this PR.

sethbern:
Thanks for the thorough review! I addressed these in the most recent pushes:

- The _catchup flag is set client-side only when replaying the stored phase, so live socket messages are unaffected. For a new session, get_peer_dashboard already resets current_phase to enableNext so stale state clears automatically.
- Switched to per-student storage keyed by assignment_id which also fixes the Copilot catch on A/B mode, where text-chat and verbal students get different phases (enableChat vs enableFaceChat) so a single course-wide key was wrong.
- Also fixed the Copilot finding on enableChat so partner rows are no longer cleared on catch-up.

bnmnetp:
I'll plan to get this merged tomorrow, that should allow for some more widespread testing.

PR #1276, opened 2026-06-30
Change stringparam to stringparams in docs

sarangjo:
In the section "Developing the Javascript for Runestone Components", the documentation references setting `stringparam` in the PreTeXt textbook, when it should be `stringparams`.


bnmnetp:
Good catch, thanks for the PR.

PR #1277, opened 2026-06-30
Fix assignment table actions cell markup

ascholerChemeketa:
A recent commit broke the cell that has the action buttons for the assignment table.


bnmnetp:
Thanks.

Weird I looked at this as both an instructor and student.  Made some changes for the student view and maybe didn't go back to verify instructor.

PR #1282, opened 2026-07-05
Jquery free components

bnmnetp:


bnmnetp:
I walked through all the pages on the sample book and it looks good to me.  Found (and fixed) just one issue after Fable had done its work.  It would be great to have someone else take this for a spin.

ascholerChemeketa:
I just built this for a local copy of my book. Tried out various complex constructs. Everything seemed to work.

PR #1293, opened 2026-07-08
Change the position of the before checklist and added one docker bullet

### LINKED ISSUE #1272: Runestone Documentation Issue: 6.2 Quick Start
**What Page were you on**
[Please copy/paste the URL from your address bar.
](https://docs.runestone.academy/en/latest/running.html)

**Describe the bug**
In section 6.2 Quick Start, the "Before running the command" checklist appears after the setup command, meaning users may run the command before completing the necessary steps. The checklist should be moved above the command. Additionally, the existing steps need two edits:

Step 1 should be split into two separate steps to clarify that users first need to set up their books directory, then navigate to the parent directory before running the command
A new step should be added reminding users to make sure Docker Desktop is running before starting

**A Screenshot is worth a thousand words**
Image

**What is your username**
@pokitkog

harrisonj2-v:
Describe the bug that was fixed
Changed the position of the bullet list before you run.

Issue number
Fixes #1272 

What is your username
harrisonj2-v

### DIFF
--- a/docs/source/running.rst
+++ b/docs/source/running.rst
@@ -51,6 +51,13 @@ The script will guide you through an interactive setup process, asking questions
 #. Navigate to where you want the configuration files created (e.g., ``mkdir ~/runestone && cd ~/runestone``)
 #. Make sure you have the prerequisites installed (see above)
 #. For Windows users: Run this from a WSL2 terminal, not PowerShell or Command Prompt
+#. Have Docker running (Docker Desktop must be started first)
+
+.. code-block:: bash
+
+   bash <(curl -fsSL https://raw.githubusercontent.com/RunestoneInteractive/rs/main/init_runestone.sh)
+
+The script will guide you through an interactive setup process, asking questions along the way to configure your Runestone server.
 
 **After setup completes:**
 


pearcej:
@harrisonj2-v This needed a prior review. Please have another student carefully try this and post what they find.

docs/source/running.rst
```diff
@@ -51,6 +51,13 @@ The script will guide you through an interactive setup process, asking questions
 #. Navigate to where you want the configuration files created (e.g., ``mkdir ~/runestone && cd ~/runestone``)
```
PuskarC, line 51: The checklist placement and Docker reminder look good. I think one requirement from issue #1272 is still missing: the original first step should be split into two separate steps, with one step for setting up the books directory and another for navigating to its parent directory before running the setup command. Right now, both actions are still combined in the same step.

PuskarC (changes requested):
The updated checklist placement and Docker reminder look good. I requested one small change to separate the books-directory setup and navigation instructions as described in issue #1272.

docs/source/running.rst
harrisonj2-v: I did not split that first bullet because it has nothing to do with the book directory, yet it's just your runestone and Docker files.

PuskarC: That makes sense. Keeping it as one step works since it refers to setting up the Runestone configuration and Docker files rather than the books directory. The checklist placement has been fixed, and the Docker reminder has also been added. The updated instructions look clear to me, and the PR looks good.

PR #1297, opened 2026-07-10
Fix: checking for multiple choice multiple answer with 10+ answers

ascholerChemeketa:
Fixes bug identified here:
https://groups.google.com/g/pretext-support/c/ISou4CfBfVw/m/piB8hKo_EgAJ?utm_medium=email&utm_source=footer

MCMA with 10+ answers stores the `givenAnswers` as strings and compare those to ints. Doesn't necessarily cause a problem on its own, but sorting with 10+ answers produces an ordering like `1, 10, 2...`.

This fixes the `givenArray` sort and also makes sure that the code does int <--> int comparisons when checking answers.

If only there was some way to specify whether something was supposed to be a number or a string... ;)

### DIFF
--- a/bases/rsptx/interactives/runestone/mchoice/js/mchoice.js
+++ b/bases/rsptx/interactives/runestone/mchoice/js/mchoice.js
@@ -457,7 +457,7 @@ export default class MultipleChoice extends RunestoneBase {
         var buttonObjs = this.optsForm.elements.group1;
         for (var i = 0; i < buttonObjs.length; i++) {
             if (buttonObjs[i].checked) {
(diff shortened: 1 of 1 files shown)


bnmnetp:
My first thought was, who makes a multiple choice with more than 10 choices!!??

I have a list for Claude, and one thing on that list is to convert the various components to TypeScript...

bnmnetp:
Ugh, this is causing 422 errors on all submitted MC answers.  The server side schema is expecting a string for the answer.

bnmnetp:
easy enough for me to fix, so don't worry about it.

PR #1298, opened 2026-07-10
PI fixes from live class testing

sethbern:
Fixes several issues noticed during live peer instruction sessions:
- Clear per-student chat catch-up state on next question / vote 2 / next
  vote so the chat panel doesn't reappear empty on the next question
- Count messages from question start, not vote-2 start, so the message
  count doesn't reset to 0 mid-question
- Show a vote 1 prompt to students, matching the existing vote 2 message
- Show assignment title on the after-class student page
- Simplify non-LLM justification submit button text to "Submit"
- Default new peer questions to Generic LLM mode when an API key is
  configured, via async_mode column default (migration included)

Note on peer.css: while testing this I found that the chat panel rendered unstyled /staticAssets/css/peer.css (from the "restore and move peer.css" commit) because the file was never added to the nginx static assets.


sethbern:
This should be good whenever you get the chance to review @bnmnetp

PR #1299, opened 2026-07-10
Add migration troubleshooting steps

### LINKED ISSUE #1268: Runestone Documentation Issue: Migration troubleshooting in 6.5
### Page found:
https://docs.runestone.academy/en/latest/running.html

## Description

While working through the installation process, I encountered a migration issue during updating in 6.5 that was resolved by manually stamping the database to the latest Alembic revision. These troubleshooting steps are not currently included in the official documentation, but could help users who continue to experience migration errors after following the standard setup process.

I suggest adding the following note to the installation or troubleshooting section of 6.5:

>If you are still experiencing migration errors, try the following commands:

>`docker compose run --rm rsmanage alembic stamp head`
 >`docker compose run --rm rsmanage alembic current`

>The second command should display the current revision with (head) next to it. If it does, the migration issue has been successfully resolved.

Including these commands would provide users with a straightforward way to recover from migration state mismatches without having to search through issue discussions or ask for support.

`GitHub: magnus-mccaslin`

PuskarC:
## Summary

- Added troubleshooting commands for migration state mismatches
- Explained how to verify that the current Alembic revision is at `head`

## Testing

- Built the documentation locally
- Confirmed the updated troubleshooting section renders correctly
-Reviewed by @harrisonj2-v and @DieuMerci225 

Fixes #1268

### DIFF
--- a/docs/source/running.rst
+++ b/docs/source/running.rst
@@ -164,6 +164,15 @@ If you used the automated setup script (``init_runestone.sh``), check the ``init
 
 For database issues, you can reset the database by running ``docker compose down -v`` to remove all volumes, then re-run the initialization steps (either manually or via ``./init_runestone.sh``).
 
+If you are still experiencing migration errors, try the following commands:
+
+.. code-block:: bash
+
+   docker compose run --rm rsmanage alembic stamp head
+   docker compose run --rm rsmanage alembic current
+
+The second command should display the current revision with ``(head)`` next to it. If it does, the migration state has been successfully updated.
+
 Developer Setup (From Cloned Repository)
 -----------------------------------------
 


harrisonj2-v:
I have reviewed this pull request

docs/source/running.rst
```diff
@@ -164,6 +164,15 @@ If you used the automated setup script (``init_runestone.sh``), check the ``init
 
 For database issues, you can reset the database by running ``docker compose down -v`` to remove all volumes, then re-run the initialization steps (either manually or via ``./init_runestone.sh``).
 
+If you are still experiencing migration errors, try the following commands:
+
+.. code-block:: bash
+
+   docker compose run --rm rsmanage alembic stamp head
```
bnmnetp, line 171: I think this is bad advice as it will just essentially tell the system that "we are all caught up" which is likely a lie...  Unless they have actually reset and reinitialized the database, then they are for sure caught up.

In general, I think migrations are best left to those running production systems.  People doing development can try to keep current, but if they get their database messed up then developers should really just reset.

PR #1311, opened 2026-07-20
FITB: Import local libraries

dbrianwalton:
The RS component code defining FITB questions had the option to define additional libraries that would be loaded when the problem is injected into the surrounding HTML. As far as I know, this had not been used beyond the default loading of `BTM` libraries. PreTeXt did not yet have any way to define those libraries.

When adding the support on the PreTeXt side, I realized that any JS files that were attempting to be loaded from the assets were in the wrong relative location, since `import()` looks relative to the calling library, which in this case is the RS component itself. So that relative locations based on PreTeXt projects is preserved, this pull request adds the document context to create an absolute URL rather than remain in a relative URL that points the wrong location.


dbrianwalton:
See https://github.com/PreTeXtBook/pretext/pull/3054

bnmnetp:
thanks @dbrianwalton

Issue #1310, opened 2026-07-20
Runestone Issue: Assignment Page Readings Numbering Off


kmcdonnell2:
**What Page were you on**
[https://runestone.academy/assignment/student/doAssignment?assignment_id=271240](https://runestone.academy/assignment/student/doAssignment?assignment_id=271240) // Assignment page for a specific assignment

**Describe the bug**
Under the readings for the assignment, the sections are off by one. Chapter 0 is labeled Chapter 1, 1 is 2, etc.  They are different from the table of content.

**A Screenshot is worth a thousand words**
<img width="630" height="623" alt="Image" src="https://github.com/user-attachments/assets/8acc4299-994a-4544-a828-1d658032a9b1" />

**Javascript Errors**
bookfuncs.js:78 No version information available for component [tex]/newcommand
(anonymous) @ bookfuncs.js:78
bookfuncs.js:78 mjvers 3.2.2
bookfuncs.js:78                mjvers 3.2
bookfuncs.js:78 Object
bookfuncs.js:78 Array(9)
bookfuncs.js:78 Timezone cookie matches, not sending timezone to server
bookfuncs.js:78 This page served by undefined
bookfuncs.js:78 triggering runestone:pre-login-complete
bookfuncs.js:78 pre-login-complete
bookfuncs.js:78 Runestone pre-login complete
bookfuncs.js:78 Loading Runestone component: activecode
bookfuncs.js:78 Loading Runestone component: shortanswer
bookfuncs.js:78 Missing template or content area for user dropdown
(anonymous) @ bookfuncs.js:78
bookfuncs.js:78 jQuery.Deferred exception: e(...).dropdown is not a function TypeError: e(...).dropdown is not a function
    at HTMLDocument.<anonymous> (https://runestone.academy/ns/books/published/Summer-2026-PD-Pathfinders/_static/prefix-runestone.f51cd4153b45055d.bundle.js:1:1298)
    at e (https://runestone.academy/staticAssets/js/jquery.js:2:30005)
    at https://runestone.academy/staticAssets/js/jquery.js:2:30307 undefined
(anonymous) @ bookfuncs.js:78
jquery.js:2 Uncaught TypeError: e(...).dropdown is not a function
    at HTMLDocument.<anonymous> (bootstrap-sphinx.js:114:27)
    at e (jquery.js:2:30005)
    at jquery.js:2:30307
bookfuncs.js:78 Waiting on MathJax!! undefined
[...]

PR #1313, opened 2026-07-21
Use PTX genreated chapter numbers when processing manifest

ascholerChemeketa:
Fix for https://github.com/RunestoneInteractive/rs/issues/1310


ascholerChemeketa:
Pushed with updates for Copilot review

Issue #1316, opened 2026-07-21
Runestone Issue: matching.js bypasses localStorageKey()


rfd131:
It appears that for matching type problems, matching.js uses its own version of checkLocalStorage() and setLocalStorage() without reference to the user email and course via localStorageKey().

We noticed responses for matching problems saving across test users in the same browser while testing PreTeXt and Runestone on our own server and diagnosed the issue via Claude Code. 

matching.js appears to be the only problem type that does not reference localStorageKey()

rfd131:
This bug seems to be along the no-server fallback and was encountered in our 'self-hosted' deployment.

bnmnetp:
Thanks, and good catch!  Easy fix.

I typically push new versions to production on the weekends, especially when there is training going on as there is this week.

PR #1317, opened 2026-07-21
Fix matching component leaking answers across users via localStorage (#1316)

rfd131:
## Problem

`MatchingProblem.checkLocalStorage()` / `setLocalStorage()` key their entry by the bare `this.divid`:

```js
localStorage.getItem(this.divid)
localStorage.setItem(this.divid, ...)
localStorage.removeItem(this.divid)
```

Every other answer-bearing component keys through the inherited `RunestoneBase.localStorageKey()`:

```js
localStorageKey() { return eBookConfig.email + ":" + eBookConfig.course + ":" + this.divid + "-given"; }
```

Because the matching key has no per-user segment, a saved matching answer is shared by every identity using the same browser. `checkLocalStorage()` is the fallback taken when there's no server answer (`repopulateFromStorage` → `checkLocalStorage`), so this is mostly latent on runestone.academy but **fully exposed on self-hosted / offline deployments** and whenever two users share a machine — student B sees student A's matching answers restored.

Detailed diagnosis in #1316.

## Fix

Use the inherited `localStorageKey()` for the three localStorage calls, exactly like every other component. Server restore (`checkServer` / `restoreAnswers`) is unaffected — only the localStorage fallback key changes.

**No backward-compatible read of the old bare-`divid` key.** That key is the defect (shared across users), so orphaning previously-saved values is intentional, not a regression.

## Testing

`matching/js/` has no existing test harness (unlike `timed`/`webwork`/`shortanswer`), so this is a source-only change, verified with `node --check`. Manually: with the fix, a matching answer saved under one identity no longer restores under a different identity in the same browser. Happy to add a test if you point me at the preferred harness.

### DIFF
--- a/bases/rsptx/interactives/runestone/matching/js/matching.js
+++ b/bases/rsptx/interactives/runestone/matching/js/matching.js
@@ -186,14 +186,14 @@ export class MatchingProblem extends RunestoneBase {
         if (this.graderactive) {
             return;
         }
-        const data = localStorage.getItem(this.divid);
+        const data = localStorage.getItem(this.localStorageKey());
         if (data) {
             const parsedData = JSON.parse(data);
             if (
                 parsedData.timestamp &&
                 parsedData.timestamp < eBookConfig.termStartDate
             ) {
-                localStorage.removeItem(this.divid);
+                localStorage.removeItem(this.localStorageKey());
                 return;
             }
             this.connections = parsedData.connections.map((conn) => ({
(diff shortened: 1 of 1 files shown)


bnmnetp:
@rfd131, @russdeforest Thanks for the PR!

There is actually a helper function `this.localStorageKey()` that constructs the key in the same way as all of the other components.  It is defined in RunestoneBase and inherited by all of the components, so you don't need to define your own helper, just use that one.

bnmnetp:
Sorry, I clicked straight to review the code and didn't fully read your explanation.  The answer field s constructed server side, as a JSON blob.  I still think using `localStorageKey` is the right way to go.

Thanks!

rfd131:
Thanks. I will re-write using localStorageKey().

PR #1318, opened 2026-07-22
Improve example solution prompt for Parsons puzzles

krishnab917:
## Prompt Engineering Update

I updated the example solution generation prompt to make the generated solutions better for Parsons puzzles.

### What worked:
- Adding more specific instructions about using beginner-level code helped prevent the model from creating overly short or complicated solutions.
- Asking for explicit loops, conditionals, and intermediate variables made the generated code easier to split into Parsons puzzle blocks.
- Adding the requirement to only output code reduced extra explanations or formatting issues.
- The new prompt consistently generated code with visible structure.
- It avoided compressed solutions that hide the reasoning process.
- It produced solutions that can naturally be divided into meaningful Parsons puzzle blocks.

### What did not work:
- The original prompt only asked for a correct solution that passed the unittest, so the model sometimes created solutions that were correct but not ideal for teaching.
- Some generated solutions used shortcuts or compressed logic that made them harder to convert into Parsons puzzle pieces.
- The new prompt may occasionally produce slightly longer code than necessary.
- Some solutions may not represent the most professional coding style, but this is acceptable because the goal is instructional clarity.

### Testing:
I tested the new prompt on several beginner programming problems, including:
- Counting values greater than 10 in an array
- Finding the largest value in an array
- Counting vowels in a string
- Calculating the average of an array
- Checking whether a number is positive

The generated solutions were correct, passed the intended requirements, and consistently used beginner-friendly structures such as loops, conditionals, and intermediate variables.

### Why I think this prompt is better:
I think this prompt is better because it balances correctness with learning. The generated solutions still need to pass the unittest, but they are now more structured, readable, and appropriate for beginner students using Parsons puzzles.

### DIFF
--- a/bases/rsptx/book_server_api/routers/personalized_parsons/get_personalized_solution.py
+++ b/bases/rsptx/book_server_api/routers/personalized_parsons/get_personalized_solution.py
@@ -244,9 +244,26 @@ def get_example_solution(api_token, language, problem_description, unittest_code
     """
 
     example_solution_user_message = f"""
-    Generate a correct example solution for the following question in {language}: {problem_description}. 
-    Here is the unittest code: {unittest_code}. The solution should be correct and pass the unittest."
-    """
+Generate a correct example solution for the following question in {language}: {problem_description}.
+
+Here is the unittest code:
+{unittest_code}
+
+The solution must be correct and pass all unittest cases.
+
+This solution will be used as the reference solution for generating Parsons puzzles for beginner programming students. Prioritize educational clarity and structural readability over minimizing lines of code.
+
+Requirements:
+- Use beginner-level {language} features only.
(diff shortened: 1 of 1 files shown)


krishnab917:
I tested the updated prompt on several beginner programming problems:
- Counting values greater than 10 in an array
- Finding the largest value in an array
- Counting vowels in a string
- Calculating an average
- Checking whether a number is positive

The new prompt consistently generated correct solutions with explicit control flow and structures that can be divided into Parsons puzzle blocks. The main improvement was avoiding overly compressed solutions while maintaining correctness.

aspadiyath:
LGTM! We can wait for Brad's review

bnmnetp (approved):

PR #1326, opened 2026-07-28
Show source repository on book metadata page

### LINKED ISSUE #732: Display repo information on author tools book metadata page
When looking at the "Edit book metadata" page in author tools for an already created book I can't tell what repo it was created from. I think I remember but it'd be nice to know for sure.

So my feature request is to put the information about what Github repo the book was configured with, even if the information is read only. Ideally I could also reconfigure it to point to a different repo but if that's hard for deep reasons at least knowing what repo it came from would be helpful.

averyquinnhq:
## Summary

- show the configured GitHub repository on the book metadata edit page
- keep the repository value read-only and show a clear fallback when none is recorded
- add rendering coverage for both cases

## Tests

- `pytest -q test/components/rsptx/templates/test_core.py` — 3 passed in an isolated environment matching the declared FastAPI 0.115 range
- `ruff check bases/rsptx/author_server_api/main.py test/components/rsptx/templates/test_core.py`
- `ruff format --check bases/rsptx/author_server_api/main.py test/components/rsptx/templates/test_core.py`

Closes #732

AI-assisted contribution prepared and tested by Avery Quinn, working with @vivid0o0.

### DIFF
--- a/bases/rsptx/author_server_api/main.py
+++ b/bases/rsptx/author_server_api/main.py
@@ -363,7 +363,13 @@ async def editlib(request: Request, book: str, user=Depends(auth_manager)):
         return RedirectResponse(url="/author/", status_code=status.HTTP_303_SEE_OTHER)
     return templates.TemplateResponse(
         "author/editlibrary.html",
-        context=dict(request=request, form=form, book=book, course=course),
+        context=dict(
+            request=request,
+            form=form,
+            book=book,
+            course=course,
+            github_url=book_data.github_url if book_data else None,
+        ),
     )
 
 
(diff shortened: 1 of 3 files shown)


bnmnetp:
Thanks!  I merged this with one small change. I moved the repo information to below all the editable data.

The only other minor improvement might be to make the repo URL clickable so that it is easy to visit.  But since it already belongs to the author that seems super low priority.

averyquinnhq:
Thanks, Bradley — moving the repository info below the editable fields makes sense. I agree the clickable-link tweak is low priority here, so I’m happy to leave it as-is. Appreciate the merge and the clear feedback.

PR #1329, opened 2026-07-30
Grader updates

ascholerChemeketa:
Two updates to the new grader:

1) A "Select all" button to allow quick selection of all the questions in an assignment if you need to regrade an entire assignment.
2) Trigger an LTI1.3 grade update message when a manual override is entered.


ascholerChemeketa:
With our LTI workflow, there should not be students in an LTI linked course without LTI1.3 user mappings... I don't think that is going to cause many failed sends.

I don't think we want to send the manual score if grades are not released (that would implicitly release the score via LMS gradebook). It will get sent when release state is changed.

bnmnetp:
Doesn't the column header for the name already allow you to select all the problems?

ascholerChemeketa:
Ahhh, didn't see that.

The default view is Cards which does not have that widget. Maybe Select All button make sense only in that view?

bnmnetp:
Yeah a button in that view would be good.

PR #1331, opened 2026-07-30
Fix: course creation - prevent comparison error for books with None for shelf_section

ascholerChemeketa:
I have a bunch of books in my dev library where shelf_section is None. That crashes the sort code when creating a new course. This fixes the course creation page to handle those.


bnmnetp:
Seems like providing a default for the shelf_section and/or making a pass through `course_list` to populate None values with Misc would be better so we don't leave books out of the library page.

Also we ought to file issues against those books that don't provide a shelf section so the author can remedy that.  I must have "fixed" all those in production a while ago.

ascholerChemeketa:
Added a default value to create_library_book that will make sure you can't add books with None for shelf_section.

Pretty sure the offending path was using addbookauthor to manually add books.

ascholerChemeketa:
Updated first commit to restrict Select All button to cards view.

bnmnetp:
I think your last comment belongs on the other PR.  Otherwise this looks good.

PR #1337, opened 2026-08-02
Improvements to developer documentation

oscarlevin:
While setting up my runestone dev environment, I got confused in section 13.  Step 6 says to leave the virtual environment.  There is then a note about how in the future commands that should happen inside the environment will include `uv run`.  But step 7 omits this.


bnmnetp:
The build command loads the `.env` file with `load_dotenv()` so they should be loaded either way.

oscarlevin:
I'll try to fix that bit.  I'm seeing some other outdated bits of the documentation.  Shall I just keep adding to changes here or would you like separate PRs?

bnmnetp:
Feel free to keep fixing in this one 😄

bnmnetp:
@oscarlevin -- this is marked as Draft so I haven't merged it...

Issue #872, opened 2025-07-08
Feature Request  Implement editor for datafiles


bnmnetp:
will provide details

morozov-av:
@bnmnetp Could I kindly ask you when would you provide the details?

morozov-av:
@bnmnetp Could I kindly ask you to review my question?

bnmnetp:
1. yes
2. no code blocks. but there is not reason why a data file could not contain code.
3. These files are saved to our badly named  `sourc_code` table.

Issue #420, opened 2023-11-03
Datafile processing order problem


bnmnetp:
In the new implementation of datafiles We should prefer the local copy of the datafile if it is on the page rather than going to the database immediately.

bnmnetp:
This is in activecode.js and should not be too hard to reorder the logic.

Issue #293, opened 2023-06-30
Testing jar files using the @datafile on java


bnmnetp:
See:  https://runestone.academy/ns/books/published/csawesome/Unit8-2DArray/pictureLabA5.html

See:  https://github.com/bhoffman0/CSAwesome/blob/master/_sources/Unit8-2DArray/pictureLabA5.rst for the rst code

rbeezer:
I've got an example together, which I will try to test on the development server in just a bit.
* Tried to "load" the jar file from an external file, which illustrates the need to parse that file as text, not XML, so there's a bug to fix before that is possible.
* Do we have enough contact info for Sydney that we can ping her on this thread?

bnmnetp:
I have her email from prior conversations.  I'll direct her to this issue.

rbeezer:
Thanks for looping in Sydney.

Very close: new output (that looks right) in
```
/home/bmiller/Runestone/books/PTXSB/published/PTXSB/data-files.html
```
but I can't view it with the incantations I have.  Could be confused about the right domain name
```
dev.runestone.academy, dev.runestoneacademy.org, 159.223.184.209
```
I know there was a change, but didn't make a note (or can't find it).

bnmnetp:
Use dev.runestoneacademy.org

bnmnetp:
In livecode.js - around line 175 we need to look at the data-filename attribute to get the real filename it does not come from the id

rbeezer:
> Rob is going to push the example in the sample book so that brad can test.

Said example just pushed at https://github.com/PreTeXtBook/pretext/commit/a1ea2b1c4799763aaf47e743d81497364e7fd997

I did not update non-RA-hosted version at PTX website, but that will likely happen in due course anyway.

bnmnetp:
Thank you,  I think I have now looked at enough code to know the way forward.

bnmnetp:
OK!  The beach example is working with the latest release on the CDN.

bnmnetp:
Closing — this was resolved at the time. The fix was in `livecode.js`, using the `data-filename` attribute rather than the id to recover the real filename, and the beach example worked with the release that followed.

Thanks @rbeezer for building the sample and pushing it into the PreTeXt sample book, which is what made it testable.

Issue #282, opened 2023-06-27
:datafile: for file input and :stdin: not working for Java/Jobe


bhoffman0:
I have not been able to get :stdin: and  :datafile: to work in Java on the JOBE server on Runestone. The Java :stdin: example here https://runestone.academy/ns/books/published/authorguide/directives/activecode.html?mode=browsing gives a server error.  The datafile that I tried is here: https://runestone.academy/ns/books/published/csawesome/Unit6-Arrays/topic-6-2-traversing-arrays.html?mode=browsing#groupwork-programming-challenge-spellchecker. It gives a server error too. Maybe it's too big of a file though.
I do use :datafile: with images in the picture lab and they work fine there.

bnmnetp:
@bhoffman0 Did the CSAwesome example work at some point?  Or was this something new?

bhoffman0:
Ah, requested entity too large. I'll try a smaller file.
<img width="179" alt="image" src="https://github.com/RunestoneInteractive/rs/assets/4571962/c7d50c5c-dc49-487d-9d67-c1a38474027e">

bnmnetp:
Wow that example is really out of date.  We have not used the jobe server in NZ for many years.

bnmnetp:
This page:  https://runestone.academy/ns/books/published/overview/ActiveCode/java.html I'm not sure but I think the temp conversion example must use stdin

bhoffman0:
Thanks! I'll compare the code for stdin and try a smaller file for datafile.

bhoffman0:
This turned out to be a bug in CodeTestHelper.java used on JOBE which was causing an exception on main methods with throws IOException for File I/O. Peter Seibel fixed the bug here https://github.com/bhoffman0/CSAwesome/blob/master/CodeTestHelper.java

bnmnetp:
OK, but the official version of the CodeTestHelper.java file is in the ~/rs repository so that if there are updates and fixes it will get built automatically when we rebuild all of our containers.  Tagging @gigamonkey

bhoffman0:
Should I do a PR? Do you need the .class file that's in there?

bnmnetp:
No class file is needed as it is just going to get recompiled by jobe when it notices that the .java file is newer.

gigamonkey:
[...]

PR #1361, opened 2026-08-11
Support binary files in ActiveCode (base64 delivery + per-language compile-also)

skiadas:
### Summary

Support for **binary files** (e.g. a compiled `.jar` or `.zip`) supplied by a PreTeXt book and made available to server-side (Jobe) ActiveCode programs.

PreTeXt's new `file` element (see the companion [pretext PR #3139](https://github.com/PreTeXtBook/pretext/pull/3139)) marks a file as binary with `data-isbinary="true"` and stores its contents base64.  These commits thread that flag through the stack so a binary file reaches the program's working directory intact, and so a program on *another* page can still recognize and use it.

### What each commit does

1. **Add is_binary to source_code** — a new `source_code.is_binary` boolean column, populated from `data-isbinary` during manifest processing, and returned by `get_source_code`.  Binary files are stored base64; text files keep `is_binary = False`.
2. **Add is_binary migration, crud, and endpoint tests** — migration round-trip plus crud and `get_source_code` endpoint coverage.
3. **Deliver binary files to the server working directory** — the `datafile` component preserves `data-isbinary` on the rendered element; `livecode` hands a binary file's base64 contents to Jobe verbatim, so the file lands in the working directory and any language can read it.
4. **Wire binary compile-also files into the build per language** — `compile-also` means different things per language: text sources compile together, a Java/Kotlin `.jar`/`.zip` goes on the classpath (`-cp`, with Jobe's Java interpreter defaults reproduced), a C/C++ `.o`/`.a` goes on the link line, and interpreted languages just read the delivered file.
5. **Add binary file tests for datafile and livecode** — vitest coverage for delivery and per-language compile-also wiring.

### Notes / limitations

- **Cross-page references** work: a program on one page can reference a `file` element living on another page, because `get_source_code` returns `is_binary` so the client treats the fetched contents correctly.
- **C/C++ shared libraries (`.so`)** are not supported: the server cannot arrange for the runtime linker to find them.  Object files (`.o`) and static archives (`.a`) work.
- **Author-facing datafile upload in the assignment builder** (`DataFilesEditor` / `instructor.py`) is **out of scope** here: it still treats uploaded files as text.  Supporting binary uploads there (reading them as base64, storing `is_binary`) is a natural follow-up.
- In-browser runtimes (e.g. Skulpt for Python, GDScript) cannot consume binary files; a binary `file` is dropped from static (non-Runestone) output by PreTeXt.

### Related

- [#872](https://github.com/RunestoneInteractive/rs/issues/872) — datafile editor feature request.  This PR adds the `source_code.is_binary` flag the author-facing editor will need; binary upload there (base64 + `is_binary`) is the natural follow-up to the out-of-scope note above.
- [#173](https://github.com/RunestoneInteractive/rs/issues/173) — "available_files does not work with languages that depend on Jobe" (2018).  This PR delivers files to the Jobe working directory for all languages and adds the cross-page DB lookup, implementing that issue's intent.
- [#282](https://github.com/RunestoneInteractive/rs/issues/282) — datafile / stdin not working for Java on Jobe.  The class of Java file-input bug this PR addresses (jar on the classpath, files in the working directory).
- [#293](https://github.com/RunestoneInteractive/rs/issues/293) — testing `.jar` files via `@datafile`.  The scenario binary-file support enables; the legacy text-`jar`/`parseJavaClasses` path is preserved.
- [#420](https://github.com/RunestoneInteractive/rs/issues/420) — prefer the on-page datafile over the DB.  Confirms the same-page-first lookup order this PR implements.

### DIFF
--- a/bases/rsptx/book_server_api/routers/rslogging.py
+++ b/bases/rsptx/book_server_api/routers/rslogging.py
@@ -665,13 +665,16 @@ async def get_source_code(
     if db_result:
         # Found data, unpack desired fields
         file_contents = db_result.main_code
+        is_binary = db_result.is_binary
         if db_result.filename:
             filename = db_result.filename
     else:
         file_contents = None
+        is_binary = None
 
     response_bundle = {
         "filename": filename,
         "file_contents": file_contents,
+        "is_binary": is_binary,
     }
     return make_json_response(detail=response_bundle)
--- a/bases/rsptx/interactives/runestone/activecode/js/livecode.js
+++ b/bases/rsptx/interactives/runestone/activecode/js/livecode.js
@@ -277,7 +277,7 @@ export default class LiveCode extends ActiveCode {
         var files = [];
         for (let f of allFilesRaw) {
             // Need to determine content and filename for each datafile
-            let fileName, content;
+            let fileName, content, isBinary = false;
             // Check on page to see if we have the datafile.
             // Datafiles are looked up via data-filename attribute while additional_files are looked up via id
             let fileElement;
@@ -310,6 +310,9 @@ export default class LiveCode extends ActiveCode {
                 // If the file came from an item with a data-filename attribute, use that as the filename
                 // otherwise this must be an RST item with filename as the id
                 fileName = fileElement.dataset.filename || f.filename;
+                isBinary =
+                    fileElement.dataset.isbinary === "true" ||
+                    fileElement.dataset.isBinary === "true";
             } else {
                 // check to see if file is in db
                 let result = null;
@@ -356,10 +359,19 @@ export default class LiveCode extends ActiveCode {
                     // favor student code if it exists
                     content = studentCode || result.file_contents;
                     fileName = result.filename;
+                    // binary files (e.g. .jar) are stored base64; the is_binary
+                    // flag tells us to hand the content to the server verbatim
+                    isBinary = result.is_binary === true;
                 }
             }
 
-            if (fileName) {
+            if (fileName && isBinary) {
+                files.push({
+                    name: fileName,
+                    content: content,
+                    isBinary: true,
+                });
+            } else if (fileName) {
                 let fileExtension = fileName.substring(
                     fileName.lastIndexOf(".") + 1,
                 );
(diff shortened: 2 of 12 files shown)


bnmnetp:
@skiadas - Thanks for this, I did see something related pop up on the PreTeXt group and as a PR?

But, I'm confused, maybe I've not spent enough time following pretext-dev. 

Does this replace datafile?  I thought that was what I read.  If so why? why not just extend what we have?  If it replaces how are we going to provide backward compatibility for all of the datafiles already in the database and in many many books?

@ascholerChemeketa would welcome your thoughts.

skiadas:
> @skiadas - Thanks for this, I did see something related pop up on the PreTeXt group and as a PR?
> 
> But, I'm confused, maybe I've not spent enough time following pretext-dev.
> 
> Does this replace datafile? I thought that was what I read. If so why? why not just extend what we have? If it replaces how are we going to provide backward compatibility for all of the datafiles already in the database and in many many books?
> 
> @ascholerChemeketa would welcome your thoughts.

Here's the relevant pretext PR: https://github.com/PreTeXtBook/pretext/pull/3139

The new file tag replaces the datafile tag in a backwards compatible way. All the old datafiles get converted to files in the preprocessor/repair, and the file tag emits the exact same manifest/html in those cases. So on the runestone end even without this PR, but just with the pretext PR, the support for text or image datafiles or (text/image files in the new format) should be still working. So that is a first test to do, grab just the pretext pr and run runestone tests on it, nothing should have changed. On the runestone side everything is still called "datafile". But now it comes from pretext files.

I just thought trying to shoehorn the new proposed syntax into the existing pretext-datafile element would be hard, confusing and misleading as these are no longer just "data" files.

What this runestone PR does allow is the the handling of the file elements where the format is not image or text but binary, like a jar file, a wheel, a zip file, a C++ object .o file etc. The pretext PR enables those file[format=binary] elements to appear in pretext source, but their proper handling falls on Runestone. 
So with this PR these would become viable as add-file entries in programs, and for example a .o file would get linked to the C++ compile step for a C++ program, or a proper jar file would be present and added to the classpath for the compilation of Java programs.

And yes I would very much like Andrew's and your eyes on both of these commits.

ascholerChemeketa:
The RS side changes seem pretty limited (binary datafile handling) and desirable regardless of what happens PreTeXt side.

Rob hasn't had a chance to weigh in on the rebooted conversation on `-dev`. So he might conceivably want some changes to the markup there, but that shouldn't affect the RS code. But in the earlier round of conversations he was broadly supportive of better handling different kinds of files authors need/want to shove into their books and providing visibility controls on the contents.

PR #1362, opened 2026-08-11
docs: propose multilingual contribution workflow

Yusuf-Gadelrab:
## Summary

Adds a docs page proposing a multilingual contribution workflow for Runestone Interactive: how translators can contribute, review conventions for RTL languages, and how course authors can flag pages that need translation.

## Testing

- Documentation-only change; `git diff --check HEAD^` clean.


bnmnetp:
I would suggest that you join the pretext-dev google group and post your text there.  I don't think you are going to get much response by adding it to our docs.

PR #1372, opened 2026-08-14
Allow browsing-mode users to get the backup Parsons problem

aspadiyath:
Previously /ns/coach/parsons_scaffolding required login unconditionally (Depends(auth_manager)), so logged-out/browsing-mode users got a hard 401 even when the exercise had a pre-authored, non-AI backup Parsons problem configured. Swap in an optional-auth dependency and, for anonymous users, serve the static backup when one exists; otherwise keep requiring login for the CodeTailor/LLM-personalized path.

### DIFF
--- a/bases/rsptx/book_server_api/routers/coach.py
+++ b/bases/rsptx/book_server_api/routers/coach.py
@@ -33,7 +33,7 @@ from fastapi import status
 from .assessment import get_question_source, SelectQRequest
 
 # Import function for fetching api - comment out for DEV purposes
-from rsptx.auth.session import auth_manager
+from rsptx.auth.session import auth_manager, NotAuthenticatedException
 from rsptx.db.crud.crud import fetch_api_token
 from rsptx.db.crud.course import fetch_course
 from rsptx.db.crud.question import fetch_question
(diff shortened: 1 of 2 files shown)


bases/rsptx/book_server_api/routers/coach.py
```diff
@@ -199,10 +199,23 @@ async def get_question_html(request: Request, div_id: str):
     return {"html": html}
 
 
+async def _get_optional_user(request: Request):
```
bnmnetp: This really should go in `components/rsptx/auth` either core.py or session.py so that it can be easily imported and reused in other endpoints.

bnmnetp: I would then remove the leading _ from the name as well.

bnmnetp (changes requested):

bnmnetp:
I was just about to start a build of all of the servers to deploy tomorrow morning.  I can hold off for a couple of hours if you think you can address my comment today.

aspadiyath:
Yes I can do that in a couple moments, thanks for holding off!

PR #1394, opened 2026-08-18
Add warnings if due dates can't be shifted when importing assignments

oscarlevin:
Courtesy of my assistant, this makes importing assignments more robust: if a due date for an assignment cannot be shifted (for example because the course timezone is not supported), then the assignment is still imported, just not with a shifted due date.

### DIFF
--- a/bases/rsptx/admin_server_api/routers/instructor.py
+++ b/bases/rsptx/admin_server_api/routers/instructor.py
@@ -1206,15 +1206,18 @@ async def copy_assignment(
                 target_course=course,
                 importing_user=user,
             )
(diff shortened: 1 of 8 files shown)


bnmnetp:
@oscarlevin -- can you fix the lint issues, and then I'll work on getting this out ASAP.

oscarlevin:
Yes, although it will not happen until this evening when I'm home from my day job. :(

bnmnetp:
OK, I may just fix it on my end (easy enough) so I can get this out.  Its creating some anxiety for the CSAwesome teachers.

oscarlevin:
Thanks, sorry I lost track of this yesterday.  Busy time of year.

PR #1404, opened 2026-08-21
Allow cross-book doenet exercises

oscarlevin:
Loading a doenet exercise that lives in another book redirected the iframe to the change course page.  This adds the "?mode=browsing" param to the iframe url so the exercise appears.  I tested it locally and seems to work.

Claude also provided a migration for existing books.  I put that on a second commit in case you don't want to deal with it and just let the books rebuild.

### DIFF
--- a/components/rsptx/build_tools/core.py
+++ b/components/rsptx/build_tools/core.py
@@ -1104,10 +1104,18 @@ def _process_single_question(
             idchild = dynamic.attrib["id"]
             qtype = dynamic.attrib["data-component"]
     if qtype == "doenet":
-        # rewrite the url in the dbtext to use the course name in the path
+        # Rewrite the url in the dbtext to use the course name (base course) in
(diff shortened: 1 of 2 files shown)


bnmnetp:
My concern with this approach is that mode=browsing  implies that the user is not logged in to the book they are looking at.  This may not have any impact on the Doenet questions and their behavior to save and restore.  

It probably doesn't since I don't think the Doenet code looks at eBookConfig, but I'd like to verify that before merging.

I think a rebuild of the book is probably sufficient since there are few enough doenet books and I can do a rebuild almost immediately after merging/releasing this PR.

bnmnetp:
Now that I've looked, and remember that we have these `*-if.html` that are loaded as part of the iframe I see that they don't even have any eBookConfig part to them, so this should work fine.

oscarlevin:
It does seem to work as expected in my testing.  And of course, the user is NOT logged into the other book in that scenario. :)

PR #1425, opened 2026-08-26
Fix CodeTailor raw Parsons markup leaking into LLM pipeline and clipboard

aspadiyath:
The "Example" field fed into the personalized-Parsons LLM pipeline was the raw pp-authoring markup ("---"-delimited blocks with #distractor/#paired tags), not compilable code. That raw text is used verbatim as the LLM prompt's [sample-solution] and, on generation/validation failure, is returned directly as the "fixed" solution -- which then gets copied to the clipboard by the codetailor "copy to clipboard" feature.

Also fix aggregate_code_to_full_Parsons_block, which merged any run of same-indentation-level lines into one block regardless of whether they were actually nested under something. This silently collapsed unrelated sibling statements (even across separate top-level statements) into a single undraggable block, most visibly for flat/sequential code where the entire example became one block. It was previously masked because the leaked raw markup's own "---" lines were incidentally reinterpreted by the frontend as block boundaries.

### DIFF
--- a/bases/rsptx/book_server_api/routers/coach.py
+++ b/bases/rsptx/book_server_api/routers/coach.py
@@ -466,7 +466,13 @@ async def parsons_scaffolding(
             "Problem Name": problem_id,
             "Problem Description": problem_description,
             "Unittest_Code": internal_test_case,
-            "Example": parsonsexample_code,  # This is the html of the example Parsons problem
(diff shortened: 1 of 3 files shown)


aspadiyath:
Added a line of code that was causing an issue with another half of the bug solved by #1416

PR #1435, opened 2026-08-26
Async PI fixes from user study feedback

sethbern:
This fixes several UI/UX issues that were identified from user studies, along with some prompt tweaking.

- The LLM peer wasn't always getting the question text or choices at all
- Prompt tweaks so the peer engages the student's actual reasoning instead of jumping straight to a scenario, and doesn't just deflect if they say they're confused
- Show a handful of saved justifications at random, each labeled with the answer it argued for on their first PI vote
- Some UI polish: theme descriptions, bigger textarea, layout fixes

### DIFF
--- a/bases/rsptx/assignment_server_api/routers/peer.py
+++ b/bases/rsptx/assignment_server_api/routers/peer.py
@@ -10,6 +10,7 @@
 # Standard library
 # ----------------
 import datetime
+import html as html_module
 import json
 import random
 import re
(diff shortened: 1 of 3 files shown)


bnmnetp:
Here is what I see after building and testing
Screenshot 2026-08-26 at 5 06 37 PM
Even after clicking "Check Me" -- infact I get the same error about currAnswer.indexOf each time I click "Check Me"

sethbern:
I think I figured out what was causing this. For previously answered questions, restoreAnswers loads that value from the server as a string but for new questions, it stays as the raw `Number` from `givenArray[0]`, which causes the error. That also explains why I did not catch this when I tested locally or during the study since the questions I tested already had stored answers.

I also addressed the copilot comments.

bnmnetp:
Ah, yes that seemingly simple change to accommodate an author that wanted more than 10 choices for an MCQ has led to more little hidden troubles.

I'll try to get this merged a bit later today.

PR #1447, opened 2026-08-28
Strengthen password validation

### LINKED ISSUE #1307: Feature Request: Stronger passwords
**Is your feature request related to a problem? Please describe.**
The website lets you have the same character 6 times in a row

**Describe the solution you'd like**
The password should require stronger characters

**Describe alternatives you've considered**
Minimum length: 12 characters (16+ is even better)
Maximum length: At least 64 characters (don't unnecessarily limit users)
Character types: Allow all printable characters, including:
Uppercase letters (A–Z)
Lowercase letters (a–z)
Numbers (0–9)
Special characters (! @ # $ % ^ & * etc.)
No common passwords: Block passwords found in known breached-password lists (e.g., "password123", "12345678")

krishnab917:
## Summary
- Increase the minimum password length from 6 to 10 characters.
- Enforce the shared password validator for instructor CSV student enrollment.
- Preserve leading and trailing password spaces while trimming other CSV fields.
- Update password validation tests for the new policy.

## Testing
- 30 password validation tests passed.
- Ruff checks passed.
- Ruff formatting check passed.
- git diff --check passed.

Closes #1307

### DIFF
--- a/bases/rsptx/admin_server_api/routers/instructor.py
+++ b/bases/rsptx/admin_server_api/routers/instructor.py
@@ -64,7 +64,11 @@ from rsptx.db.crud import (
 from rsptx.auth.session import auth_manager
 from rsptx.auth.email import send_welcome_email
 from rsptx.templates import get_shared_templates
-from rsptx.validation.fields import clean_text, validate_text_field
+from rsptx.validation.fields import (
+    clean_text,
+    validate_password,
+    validate_text_field,
+)
 from rsptx.validation.schemas import AssignmentsSearchRequest
 from rsptx.configuration import settings
 from rsptx.endpoint_validators import with_course, instructor_role_required
@@ -1047,7 +1051,9 @@ async def enroll_students(
         # Spreadsheet exports routinely carry stray whitespace -- most painfully a
         # newline inside a cell, which used to be stored verbatim and then break
         # eBookConfig (and therefore every component) on the student's pages.
-        row = [field.strip() for field in row]
+        row = [
+            field.strip() if index != 4 else field for index, field in enumerate(row)
+        ]
         if not any(row):
             # A blank line, e.g. the trailing newline at the end of the file.
             continue
@@ -1075,6 +1081,17 @@ async def enroll_students(
             row = row[:6]  # Only take the first 6 columns
             mess = "Row has more than 6 columns, truncating."
             rslogger.warning(f"Row has more than 6 columns, truncating: {row}")
+        password_error = validate_password(row[4])
+        if password_error:
+            results.append(
+                {
+                    "username": row[0],
+                    "status": f"error: {password_error}",
+                    "category": "error",
+                }
+            )
+            failed += 1
+            continue
         try:
             user_data = AuthUserValidator(
                 username=row[0],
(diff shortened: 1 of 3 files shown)


krishnab917:
I wanted to provide a more detailed explanation of the implementation and the reasoning behind the changes in this PR.

What changed

This PR strengthens Runestone’s password policy while keeping the existing validation architecture intact.

1. Increased the minimum password length from 6 to 10

The existing validate_password() helper in components/rsptx/validation/fields.py accepted passwords with a minimum length of 6 characters.

I changed the default minimum to 10 characters:

def validate_password(password: str, min_length: int = 10) -> str | None:

I kept this as a default parameter rather than hard-coding the value throughout the application so that the password policy remains centralized in one validation helper.

The existing whitespace-only protection remains unchanged. A password containing only whitespace is still rejected, while legitimate spaces remain allowed.

2. Kept the password character policy permissive

I did not introduce requirements such as “must contain an uppercase letter, lowercase letter, number, and special character.”

The current validator already treats the password as an opaque string and only validates whether it is non-empty/meaningful and meets the minimum length. I preserved that behavior so that users can use passphrases and other strong password formats without imposing arbitrary character-class requirements.

There is also no new password maximum-length restriction in this PR. This avoids unnecessarily limiting long passphrases and is consistent with the intent of the issue to support passwords of at least 64 characters.

3. Applied the same validation to instructor CSV enrollment

The instructor interface allows students to be registered through a CSV containing:

username,email,first_name,last_name,password,course

The enroll_students endpoint in bases/rsptx/admin_server_api/routers/instructor.py previously created users from this CSV without calling the shared password validator.

That meant the regular registration flow and instructor-created accounts could have different password policies.

This PR imports the existing validate_password() helper into the instructor router and validates row[4], which is the password field, before creating a new user.

When a password is invalid, the row is rejected and reported through the existing CSV enrollment results mechanism rather than creating the account.

4. Preserved password whitespace during CSV processing

While tracing the CSV path, I also found that the entire CSV row was previously processed with:

row = [field.strip() for field in row]

That unintentionally stripped leading and trailing whitespace from passwords.

The password validator explicitly does not strip passwords because spaces can legitimately be part of a password/passphrase. I therefore changed the CSV normalization so that all fields except the password are stripped:

row = [
    field.strip() if index != 4 else field
    for index, field in enumerate(row)
]

This preserves the exact password supplied in the CSV while retaining the existing whitespace cleanup for usernames, names, email addresses, and the course field.

5. Updated the test coverage

The password validation tests in test/components/rsptx/validation/test_fields.py were updated to reflect the new policy.

The tests now cover:

Passwords shorter than 10 characters being rejected.
A password of exactly 10 characters being accepted.
Longer passphrases being accepted.
Passwords containing numbers and special characters being accepted.
Passwords containing leading/trailing spaces being accepted.
Whitespace-only passwords continuing to be rejected.

This keeps the tests focused on the actual policy rather than enforcing arbitrary password composition rules.

Validation performed

I ran the following checks locally:

uv run pytest test/components/rsptx/validation/test_fields.py
30 tests passed
uv run ruff check ...
All checks passed
uv run ruff format --check ...
3 files already formatted
git diff --check
Passed
Python syntax compilation of the modified instructor router
Passed

The PR's GitHub Actions checks are also both passing:

Lint / lint: Done
Tests / crud-tests: Done
Scope / existing accounts

PR #1448, opened 2026-08-29
Matching: override PTX SVG background color

ascholerChemeketa:
Fix for issue reported by Sean for matching questions in PTX dark mode.


bnmnetp:
Dark / Light mode feels like such a game of whack-a-mole!  I would swear that when I merged the PR to update matching that both looked good on that day at that time!

PR #1457, opened 2026-09-01
Report assignment-download timestamps in the instructor's local time

dylanpulver:
`download_assignment` adds the timezone offset where every other reader subtracts it, so the CSV lands **two offsets away** from the truth: for a US Eastern instructor a 02:00 UTC submission is written as `07:00` instead of `21:00` the day before.

`tz_offset` is `Date.getTimezoneOffset() / 60` from `bookfuncs.js`, i.e. UTC minus local (`+5` EST, `-5.5` IST), and `useinfo.timestamp` is naive UTC from `canonical_utcnow()`. So local time is the stored timestamp *minus* the offset — what `analytics.py` does at all ten of its conversion sites (`x - tdoff`) and what `rsptx/practice/core.py:39` does. `instructor.py` is the one site in eleven that adds, so the downloaded logfile disagrees with the Chapter Activity and Assignment Overview reports the same instructor sees in the same session.

Two smaller defects on the same four lines:

- `int(tzoffset)` is the only `int` cast among the six readers of this field; the rest use `float(...)` and the schema declares it `float`. Half-hour zones (India, South Australia, Newfoundland) silently lose 30 minutes.
- `parsed_js` is assigned inside `if RS_info:` but read unconditionally below it, so an instructor with no `RS_info` cookie gets an `UnboundLocalError`. It is the only one of nine inline `json.loads(RS_info)` sites with no initializer.

The patch adopts the block `analytics.py` already uses, which closes all three.

**Tests.** Four cases asserting the timestamp in the CSV rather than the offset: EST, IST, UTC, no cookie. Three fail on current `main` (`07:00:00`, `21:00:00`, a crash); the UTC case passes in both trees. Fixing only the cast leaves both offset cases red, only the sign leaves the half-hour case red, both leaves the no-cookie case red. Full suite 553 before, 557 after. Nothing covered this route previously — the file stops after `duplicate`/`delete`.

Not done here: the conversion is hand-written at eleven sites and `json.loads(RS_info)` at nine, with no shared helper, which is how a new report can re-derive the rule backwards. A `tz_offset_from_cookie()` / `to_local()` pair in `rsptx.response_helpers.core` would close that, but it spans two servers. Happy to follow up if you want it.

### DIFF
--- a/bases/rsptx/assignment_server_api/routers/instructor.py
+++ b/bases/rsptx/assignment_server_api/routers/instructor.py
@@ -1514,22 +1514,25 @@ async def do_download_assignment(
             detail=f"Assignment questions for {assignment_id} not found",
         )
 
+    # ``tz_offset`` is UTC minus local time in hours (Date.getTimezoneOffset() / 60),
+    # and useinfo timestamps are stored as naive UTC, so local time is the stored
+    # timestamp *minus* the offset -- as in analytics.py and rsptx.practice.core.
+    tz_offset_hours = 0.0
     if RS_info:
         rslogger.debug(f"RS_info Cookie {RS_info}")
         # Note that to get to the value of the cookie you must use ``.value``
         try:
-            parsed_js = json.loads(RS_info)
+            tz_offset_hours = float(json.loads(RS_info).get("tz_offset", 0))
         except Exception:
-            parsed_js = {}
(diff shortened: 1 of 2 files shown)


bnmnetp:
Thanks for the PR.
I'll get this merged when I'm back from vacation.

PR #1468, opened 2026-09-03
Ensure webwork js on assignment pages even using pretext templates

oscarlevin:
Currently if a book doesn't contain any WeBWorK questions and the instructor has the "Use PreTeXt based student pages" selected, there is no pretext-webwork.js loaded on the doAssignment page.  This PR fixes that (as well as the resizer which is also missing).  Assignments are checked for whether they contain any WeBWorK questions, and if so, we add the additional js (if it isn't already loaded).

### DIFF
--- a/components/rsptx/templates/assignment/student/doAssignment.html
+++ b/components/rsptx/templates/assignment/student/doAssignment.html
@@ -14,6 +14,31 @@ Do Assignment
     {% if origin == "PreTeXt" and using_ptx_base != "true" %}
     <link href="/ns/books/published/{{ course.base_course }}/_static/pretext/css/theme.css" rel="stylesheet" />
     {% endif %}
+
+    {# A book built without WeBWorK does not load pretext-webwork.js in its
+       generated _base.html, so an assignment that borrows WeBWorK questions
+       from another book has no handleWW() to activate them.  Load the support
+       here when the assignment needs it and nothing upstream supplied it.
+       handleWW() is only reached from the Activate button's onclick, so the
+       injected scripts have until the first click to arrive. #}
+    {% if needs_webwork %}
+    <script>
+        (function () {
+            var load = function (src) {
+                var s = document.createElement("script");
+                s.src = src;
+                document.head.appendChild(s);
+            };
+            if (typeof handleWW === "undefined") {
+                load("/ns/books/published/{{ course.course_name }}/_static/pretext/js/dist/pretext-webwork.js");
+            }
+            // handleWW() calls iFrameResize() on the problem iframe.
+            if (typeof iFrameResize === "undefined") {
+                load("https://webwork.runestone.academy/webwork2_files/node_modules/iframe-resizer/js/iframeResizer.min.js");
+            }
+        })();
+    </script>
+    {% endif %}
 {% endblock %}
 
 
--- a/components/rsptx/templates/assignment/student/doAssignment.html
+++ b/components/rsptx/templates/assignment/student/doAssignment.html
@@ -24,17 +24,23 @@ Do Assignment
     {% if needs_webwork %}
     <script>
         (function () {
+            // A dynamically inserted script defaults to async, so two of them
+            // execute in whatever order they finish downloading.  Setting
+            // async = false puts them on the in-order list instead, which is
+            // what keeps the pair below in dependency order.
             var load = function (src) {
                 var s = document.createElement("script");
+                s.async = false;
                 s.src = src;
                 document.head.appendChild(s);
             };
-            if (typeof handleWW === "undefined") {
-                load("/ns/books/published/{{ course.course_name }}/_static/pretext/js/dist/pretext-webwork.js");
-            }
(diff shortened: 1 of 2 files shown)


components/rsptx/templates/assignment/student/doAssignment.html
```diff
@@ -14,6 +14,37 @@
...
+            // The resizer goes first: handleWW() calls iFrameResize() on the
+            // iframe it builds for the problem.
+            if (typeof iFrameResize === "undefined") {
+                load("https://webwork.runestone.academy/webwork2_files/node_modules/iframe-resizer/js/iframeResizer.min.js");
+            }
+            if (typeof handleWW === "undefined") {
+                load("/ns/books/published/{{ course.course_name }}/_static/pretext/js/dist/pretext-webwork.js");
+            }
+        })();
+    </script>
```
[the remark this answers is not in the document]
oscarlevin: I do not think we need to be as careful as copilot suggests here.  If the user clicks activate and handleWW hasn't finished loading, they can just click it again.

oscarlevin:
Note also that webwork in a reading assignment would not be displayed on this page anyway, so that is not a concern.

components/rsptx/templates/assignment/student/doAssignment.html
```diff
@@ -14,6 +14,37 @@
...
+            var load = function (src) {
+                var s = document.createElement("script");
+                s.async = false;
+                s.src = src;
+                document.head.appendChild(s);
+            };
+            // The resizer goes first: handleWW() calls iFrameResize() on the
+            // iframe it builds for the problem.
+            if (typeof iFrameResize === "undefined") {
+                load("https://webwork.runestone.academy/webwork2_files/node_modules/iframe-resizer/js/iframeResizer.min.js");
```
bnmnetp, line 40: Is this missing from the /ns/books/published/.... path? It feels like it should be there for WW questions in context.  I don't recall loading any other resources from webwork.runestone.academy

oscarlevin: Looking at live assignment pages that do have webwork, this is loaded from whatever webwork server the book is using.  I agree that is strange since PreTeXt does ship some sort of iframeresizer.  

Oh, actually it is lti_iframe_resizer.js that comes with pretext.  So I think that is different.

ascholerChemeketa:
Going to go on the grinch record here with this take:

If a book doesn't use some external tool, maybe it should not be available in Runestone assignment builder for that book.

Should Runestone inject Doenet js? How about stack? How about Asymptote in case the user wants to add their own 3D diagram? Where is the line?

Even if that take is too hot, having Runestone responsible for multi-step loading sequences and timing stubs to load other tools seems like something that we would like to avoid. More reasonable would be one "load webwork" and let it be responsible for avoiding double initialization, loading other needed components (resizer), etc...

If that take is still too hot, has this been tested?

oscarlevin:
> If that take is still too hot, has this been tested?

Yes, tested with both books using and not using webwork in both pretext-student-pages and classic, as well as the assignment builder interface.


> If a book doesn't use some external tool, maybe it should not be available in Runestone assignment builder for that book.
> 
> Should Runestone inject Doenet js? How about stack? How about Asymptote in case the user wants to add their own 3D diagram? Where is the line?

I think the line is right after: are exercises in the runestone database available?  If Runestone has a database of exercises, then those should work in runestone.  It shouldn't matter which book you have set as your base course.  Otherwise, why not just say, if a book doesn't have an exercise in its source, then it shouldn't be available to that course.

Collection of exercises (exercise banks) make runestone a lot more appealing.  But that only works if they can be shared.

> Even if that take is too hot, having Runestone responsible for multi-step loading sequences and timing stubs to load other tools seems like something that we would like to avoid. More reasonable would be one "load webwork" and let it be responsible for avoiding double initialization, loading other needed components (resizer), etc...

Perhaps the correct solution here is that pretext should always load the webwork javascript for all runestone builds.  I guess that is how the pretext based student pages get their wrapper?  I would also be open to better modularization than what this PR has, but since the current state breaks the courses for a number of current users, I went with this.

ascholerChemeketa:
> I think the line is right after: are exercises in the runestone database available? If Runestone has a database of exercises, then those should work in runestone. It shouldn't matter which book you have set as your base course.

It does in that the base course is defined by the book. Which controls what libraries and other resources are available/required in the context of that book.

The issue goes far beyond webwork. Should every book's assignment page include every possible library for every question? Should we inject the 30MB WASM runtime for GDScript into every book in case an instructor wants to use a question that involves it? I have some custom JS libraries that are used for questions in my book. Should we inject them everywhere in case someone tries to borrow a question? (How would we even know that they exist?) What about questions that rely on datafiles or the code that are defined in one book? ...

Write once, assign anywhere might be nice. But the logical way to make that possible would be to make questions self contained. i.e. the webwork question knows how to load the right libraries if they are not already available. That way every book is not responsible for dealing with every potential question type.

PR #1520, opened 2026-09-17
Retire flagged exercises instead of deleting them

oscarlevin:
## The problem

The editorial page gives an editor two choices for a flagged question: clear the flag, or **delete** it. `delete_question` runs a real `DELETE FROM questions`, and three tables reference `questions.id` with `ON DELETE CASCADE`:

- `assignment_questions` (`models.py:697`)
- `question_tags` (`models.py:633`)
- `question_grades` (`models.py:931`)

So one click removes the exercise from **every assignment in every course still using it**, along with its points and grading configuration. `useinfo.div_id` has no foreign key, so the student answer log survives as orphaned rows pointing at a question that no longer exists.

A course can lose an assignment question mid-term, and neither the instructor nor the students get any warning. An editor triaging a flagged exercise has no way to see who depends on it.

## What this changes

Retiring instead of deleting. `questions.retired_on` is stamped (with `retired_by`), the row stays, and the exercise drops out of the three instructor-facing search paths:

- `search_exercises` — the assignment builder's exercise search
- `fetch_questions_by_search_criteria` — `/search_questions`
- `fetch_questions_for_chapter_subchapter` — the book-browsing chooser (`fetch_chooser_data`)

Courses that already assign the exercise are untouched. What changes is that nobody can build a *new* assignment around it.

**Deliberately not filtered**, because these are how a live course reaches the exercise: `fetch_question`, `fetch_question_by_id`, `fetch_matching_questions` (selectquestion), `fetch_qualified_questions` (practice), and the page-progress query in `crud/book.py`. Retiring is invisible to students.

The editorial page's Delete button becomes Retire; there is no hard-delete path left in the UI.

## Why a timestamp rather than a boolean

So the deletion can still happen later, once it is provably safe. `rsmanage questions purge` deletes exercises that cleared two windows: retired longer than `--retired-for-days` (default 365), and untouched for `--unused-for-years` (default 3) by any of four signals:

- an `assignment_questions` row in a course whose `term_start_date` falls in the window
- a `useinfo` row for the `div_id` — catches reading-page exercises that were never formally assigned
- a row in any registered answer table — iterates `runestone_component_dict`, so a new `@register_answer_table` type is covered automatically
- `from_source = True` — still compiled into the book, so a rebuild would just recreate it

Every check is batched over the whole candidate set rather than run per question; `useinfo` is far too large to probe a row at a time.

There is **no override flag**. An exercise still tied to anything is kept and the report says why. It is a dry run unless given `--apply`.

```
rsmanage questions listretired
rsmanage questions retire <name> <base_course>
rsmanage questions unretire <name> <base_course>
rsmanage questions purge [--retired-for-days N] [--unused-for-years N] [--base-course X] [--apply]
```

## Migration

`c3f7e2a8b491` adds `retired_on` (nullable `DateTime`) and `retired_by` (`String(512)`), plus a partial index on the retired rows only — every search path tests `retired_on IS NULL`, which matches nearly every row, so a plain b-tree index would never be chosen; the partial index is small and is what the purge scans.

Single head after this, on top of `f3b8d5c2a710`.

## Testing

`test_editor_routes.py` is updated for the new endpoint and adds `test_retired_question_drops_out_of_search`, which asserts the exercise is returned by `search_exercises` before retiring and absent after.

```
test/bases/rsptx/admin_server_api/test_editor_routes.py   10 passed
test/components/rsptx/db                                 123 passed
test/bases/rsptx/assignment_server_api                   127 passed
```

Also exercised by hand in a full docker compose stack: retire from the editorial page, confirm the exercise disappears from both exercise search and the book chooser while the existing assignment using it keeps working.

## Note for deployers

Polylith bundles a copy of `components/` into each project wheel, so this change needs `admin` (endpoint + template), `assignment` (all three search filters run there), and `rsmanage` (the purge command) rebuilt. It also touches `components/rsptx/templates/staticAssets/js/admin/manage_exercises.js`, which the proxy serves from a copy baked into its own image — so `caddy`/`nginx` needs rebuilding too, or the Retire button throws `retireQuestion is not defined` against a cached script. `book_server` does not need rebuilding; none of the filtered functions are called from it.


bnmnetp:
@oscarlevin -- maybe hold off a bit on this until I merge #1537

PR #1522, opened 2026-09-17
Pass CADDY_SITE_ADDRESS and CERTBOT_EMAIL from .env to admin server

ascholerChemeketa:
Makes sure that we pass settings that the auth logic may inspect from the `.env` file into the admin container.  (Failure to do so was causing a cookie issue in my admin server.)

Side note: I think there is a broader need to audit what is getting passed in, or possible always pass the entire `.env` contents? Sean recently ran into the fact that `landing_page_url` never makes it through from .env into a container. I think there are other settings that are being ignored. Would it make sense to always pass all the env variables though for all containers instead of manually unpacking the desired ones?


bnmnetp:
It certainly broadens our security risk to have every container have everything.  Maybe there is a core set like (assignment, book, admin) that should all get an identical set.  Or I just need to be more careful about making sure new ones are added in all the appropriate places.

ascholerChemeketa:
@bnmnetp What do you think about the CADDY_SITE_ADDRESS issue flagged? It feels like the real issue is in the server_protocol prop:
```
        if self.caddy_site_address:
            return "https://"
``` 

Should we change that to something like ` if self.caddy_site_address and self.caddy_site_address.startswith("https:")`?

bnmnetp:
Probably, although in practice I'm not sure anyone would use :80 ??

As I was thinking about this I remembered why I had not added `LANDING_PAGE_URL` and even this `CADDY_SITE_ADDRESS`.  My goal is for people like Sean to use the `docker-compose.override.yml` file for customizing things they need when running their own public facing server.

So maybe I need to do another pass over the docs and to be sure we have them complete and document those that should be set in the override file when running your own server.

ascholerChemeketa:
Maybe... though it feels weird to have to customize the setting in two places: both the `.env` and then in `docker-compose.override.yml`.  

`docker-compose.override.yml` feels like it should be for completely changing containers (e.g. I use it to remap the caddy servers' staticAssets to the source dir for live updates), not setting saying pass this environmental variable that I set in .env through to code that is looking to see if it is set.

Or, if you want to move to docker-compose as the customization spot, maybe those things should move completely out of .env???

bnmnetp:
But we already do that for every environment variable in docker-compose.yml We don't want to be exposing private information, and I don't want people to have to mess with merging their changes to a docker-compose.yml file with ours.  That is exactly the purpose of the .override file.

Maybe things that everyone would agree are "non-private" should just live in the docker-compose.yml file with the appropriate defaults.


Claude just suggested that we restructure the docker-compose.yml file like this:

```
# ---- reusable environment fragments -------------------------------
x-db-env: &db-env
  DEV_DBURL: ${DC_DEV_DBURL:-$DEV_DBURL}
  DBURL: ${DC_DBURL:-$DBURL}
  DB_POOL_SIZE: ${DB_POOL_SIZE:-10}
  DB_MAX_OVERFLOW: ${DB_MAX_OVERFLOW:-20}
  DB_POOL_TIMEOUT: ${DB_POOL_TIMEOUT:-10}
  DB_POOL_RECYCLE: ${DB_POOL_RECYCLE:-1800}
  DB_SYNC_POOL_SIZE: ${DB_SYNC_POOL_SIZE:-5}
  DB_SYNC_MAX_OVERFLOW: ${DB_SYNC_MAX_OVERFLOW:-10}

x-uvicorn-env: &uvicorn-env
  UVICORN_USER: www-data
  UVICORN_GROUP: www-data
  UVICORN_MAX_REQUESTS: "5000"
  UVICORN_MAX_REQUESTS_JITTER: "30"
  UVICORN_TIMEOUT: "60"

x-lti-env: &lti-env
  LTI1P3_PUBLIC_KEY: ${LTI1P3_PUBLIC_KEY}
  LTI1P3_PRIVATE_KEY: ${LTI1P3_PRIVATE_KEY}

x-telemetry-env: &telemetry-env
  TELEMETRY_ENABLED: ${TELEMETRY_ENABLED:-true}
  TELEMETRY_URL: ${TELEMETRY_URL:-https://runestone.academy/admin/telemetry/checkin}
  TELEMETRY_REGION: ${TELEMETRY_REGION:-}
  TELEMETRY_INSTITUTION: ${TELEMETRY_INSTITUTION:-}

# shared by book / assignment / admin
x-app-env: &app-env
  BOOK_PATH: /usr/books
  SERVER_CONFIG: ${SERVER_CONFIG}
  RUNESTONE_PATH: /usr/src/app
  REDIS_URI: redis://${REDIS_HOST}:6379/0
  DOCKER_COMPOSE: "1"
  JWT_SECRET: ${JWT_SECRET}
  LOG_LEVEL: ${LOG_LEVEL:-INFO}
  AUTHOR_SERVER_URL: ${AUTHOR_SERVER_URL:-https://author.runestone.academy/author}
```

Then each service keeps only what's genuinely its own:

```
  assignment:
    environment:
      <<: [*app-env, *db-env, *uvicorn-env, *lti-env]
      UVICORN_WORKERS: "3"

  book:
    environment:
      <<: [*app-env, *db-env, *uvicorn-env, *lti-env, *telemetry-env]
      RUNESTONE_HOST: ${RUNESTONE_HOST}
      LOAD_BALANCER_HOST: ${LOAD_BALANCER_HOST}
      LOGIN_URL: /admin/auth/login
      UVICORN_WORKERS: ${UVICORN_WORKERS:-2}
      FERNET_SECRET: ${FERNET_SECRET}
      SPACES_KEY: ${SPACES_KEY}
      SPACES_SECRET: ${SPACES_SECRET}
``` 

That seems cleaner and reduces all the duplicated variables.

ascholerChemeketa:
That looks good in terms of simplifying the docker-compose.

Anything that helps eliminate "updated .env but didn't also update docker-compose.overide" as a failure option for users. Or, "added a settings check, forgot to make sure all the containers get that value passed from .env into the container" as a failure option for devs.

PR #1527, opened 2026-09-20
LTI1p3: drop detailed logging to DEBUG level

ascholerChemeketa:
Drops the new detailed logging for LTI1p3 to DEBUG level. That should keep it out of production but keep it available in dev instances.


ascholerChemeketa:
Fixed the unused `raw_body` via force push. Pretty sure there is no way to have None in those flagged spots in practice.

